PR #6, opened 2021-04-01
Move the default cylc-run directory location

jjguerrette:
Updated README and drive.csh for a new default cylc-run location.  Need to modify ~/.cylc/global.rc in order to keep using this Repo.


junmeiban (approved):
Tested. It works! Thank you so much for adding this function.

PR #16, opened 2021-08-12
New environment and build associated with that environment

jjguerrette:
Introduces the new environment setup from https://github.com/JCSDA-internal/mpas-bundle/pull/106.

`config/environment.csh` in develop will not work with new builds.  This new environment is not backward compatible with old code versions.  Anybody using an old code version will need to use the old environment.

Also `drive.csh` is modified to limit the maximum number of active cycle points.  When many cycle points are active in complex workflows (i.e., eda_3denvar or "CriticalPathType==Bypass"), MPAS-Workflow consumes large quantities of login node resources.  This change reduces the maximum to 4 active cycle points (date-time combinations).

### DIFF
--- a/config/builds.csh
+++ b/config/builds.csh
@@ -15,7 +15,7 @@ setenv BuildCompiler 'gnu-openmpi'
 # Note: at this time, all executables should be built in the same environment, one that is
 # consistent with config/environment.csh
 
-set commonBuild = /glade/work/guerrett/pandac/build/mpas-bundle_gnu-openmpi_bugfix--out-nchans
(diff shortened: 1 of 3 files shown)


jjguerrette:
Added a comment in PR:

Also `drive.csh` is modified to limit the maximum number of active cycle points.  When many cycle points are active in complex workflows (i.e., eda_3denvar or "CriticalPathType==Bypass"), MPAS-Workflow consumes large quantities of login node resources.  This change reduces the maximum to 4 active cycle points (date-time combinations).

jjguerrette:
@metdyn, I'm going to merge this before noon today in the interest of having a functional develop branch.  The mods work for me.  If you feel something is wrong with the changes, we can make other improvements afterward.

PR #18, opened 2021-09-23
Fix clean tasks

jjguerrette:
`Clean*` tasks were previously run in the background on the same node as where `drive.csh` is executed.  That causes lagging when deleting large numbers of files simultaneously, i.e., when cleaning up from verification of large numbers of ensemble members simultaneously.

Rightfully, that makes HPC managers very unhappy, and can result in being kicked off when drive.csh is run from a login node.  Now those cleaning tasks are submitted to single processors on the `share` queue.

Also there was a missing `source config/tools.csh` in `clean-hofx.csh`, and a `-q` flag to grep was causing background files to be duplicated unnecessarily. Those are fixed.

### DIFF
--- a/clean-hofx.csh
+++ b/clean-hofx.csh
@@ -37,6 +37,7 @@ endif
 # =================
 source config/experiment.csh
 source config/filestructure.csh
+source config/tools.csh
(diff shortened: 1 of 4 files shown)


ibanos90 (approved):
Tested the changes and all running successfully!

PR #19, opened 2021-09-24
Delete zeroth forecast and diag files

jjguerrette:
The zero duration forecast and diagnostics files from the MPAS-Atmosphere integration are not presently used in cycling.  They account for roughly half the disk space in the CyclingFC directory, which is substantial as the number of ensemble members grows.  This PR removes those unneeded files.

The zero duration forecast files from extended forecasts are used for 0-hr (analysis state) verification, and so cannot be deleted.  However, they account for relatively less disk space.


ibanos90 (approved):
Tested the changes and the size of each cycle in the CyclingFC directory is reduced by half (for example, from 17G to 8.5G). The zeroth forecast remains in the cycles folders under the ExtendedFC directory.

jjguerrette:
Thank you @ibanos90!

Issue #20, opened 2021-09-24
Separate jediPrep task out of CyclingDA task instead of performing as env-script


jjguerrette:
jediPrepCyclingDA takes a very long time as the number of EDA members increases.
- [x] jediPrepCyclingDA should be a separate cylc task in the share queue instead of using the same number of processors as CyclingDA.  That will significantly reduce computational overhead for EDA cycling. Closed by #29.
- [x] Create variationalPrep.csh, which moves the inner/outer loop and Ensemble B sed substitution components from jediPrep.csh into a dedicated script for variational applications. Addressed by #31.
- [x] Move the background, static, and template file preparation from variational.csh into variationalPrep.csh in order to run it on a share queue and avoid extra time being spent within the large footprint CyclingDA task. Addressed by #32.
- [x] Ensemble B preparation can be sped up for EDA by more smartly generating the individual Ensemble B member yaml entries in the individual EDA member_N.yaml files (currently resides at the end of `jediPrep.csh`).  Too many repeated `cat` operations and maybe too much `sed` is used now.  Those can be replaced by a much easier-to-follow python script.

PR #31, opened 2021-10-14
Create new variationalPrep.csh for VariationalIterations and Ensemble Pb, improve StaticFieldsFile handling

jjguerrette:
7183787 and d7688c7
Some effort is made to reduce the EDA wall-time caused by concurrent access (across EDA members) of a single static.nc file via the static stream.  Now each EDA member gets its own static file.  There is not much impact to the wall-time, but this is no longer an unknown in terms of performance.

c93512e
The YAML sed substitution for VariationalIterations and all Ensemble B components is moved from jediPrep.csh (used for both hofx and variational) to variationalPrep.csh.

Addresses the second bullet in #20.

The sed substitution for `AppType` (i.e., `hofx` or `variational`) in the observation feedback file names is removed from the `ObsPlugs` and instead included in the file prefix substitution within `jediPrep.csh`.  The goal is to improve transparency in file naming conventions.


ibanos90 (approved):
Tested for eda_3denvar and 3denvar and all changes are working well.

PR #32, opened 2021-10-14
Independent EDA member jobs

### LINKED ISSUE #20: Separate jediPrep task out of CyclingDA task instead of performing as env-script

### LINKED ISSUE #30: Create a new option to run eda_3denvar Variational members independently
### Basic functionality of the change
The new default will be to run EDA members independently, because running all simultaneously requires significant computational resources.

Keep the option to run EDA members simultaneously in single executable for future investigations of Block EDA methods, which can be carried out at lower resolution.

### Motivation for the record
The large cost of the single-executable EDA is due to (1) the significant number of requested nodes (causes very long queue wait times) and (2) load imbalance between EDA members.  (1) is the dominating factor.  (2) causes the single-executable EDA to be on the order of 30-80% more expensive in terms of core-hours than independently executed variational applications with the same number of processors per member.  Although memory availability may be a contributing factor to (2), which I am unable to discern at this time.

For example, when running an 80-member EDA of 3denvar's on the 120km mesh, the fastest member at 2018041506 takes ~600 sec. on 96 nodes x 30 pe/node, and the slowest member takes ~900 sec.  The aggregate load imbalance is around 38% across EDA members.  With only 96 nodes, and 36 processor per member, all members are split unequally across 2 nodes with the following repeating pattern for every group of 5 members:
30 6
24 12
18 18
12 24
6 30

The load imbalance occurs across a broad set of class methods and is not restricted to the I/O parts of the executable.  In particular, the `Localization::multiply` method uses ~50% of the wall-time and has a load imbalance of ~60%.  It is possible that the large load imbalances across the whole EDA application are indicative of a memory bottleneck.  The maximum memory load used (all processes) is 3989.8 GB and the total memory available is 45GB * 96 nodes = 4320 GB.  It's difficult to say whether the slower-running members were more memory-limited than the faster-running members.  It is possible that the node sharing between members is causing competitions for memory.  I think that I would need member-by-member timing and memory usage information in order to determine what is the underlying cause of the aggregate load imbalance between members.

jjguerrette:
### Commit descriptions
096757f
Changes from using relative directories to absolute directories in namelist.atmopshere, streams.atmosphere, and most YAML files.  The purpose is to allow for applications (only Variational for now) to be run in a different directory than the run directory.  This facilitates extending the Variational application to an ensemble of running tasks more simple.

57109f0
Several scripts are renamed for clarity, and the changes are cascaded through the rest of the workflow
 - `clean-hofx.csh` -> `CleanHofX.csh`
 - `clean-variational.csh` -> `CleanVariational.csh`
 - `hofx.csh` -> `HofX.csh`
 - `jediPrep.csh` -> `PrepJEDI.csh`
 - `variationalPrep.csh` -> `PrepVariational.csh`
 - `variational.csh` -> `Variational.csh`

cc33314
Addresses third bullet of #20.
Move background state preparation from `Variational` (formerly `variational`) to `PrepVariational` (formerly `variationalPrep`) to reduce wall-time in CyclingDA task.

e2b0799
Very small clarification to a comment

2ca1535
Small refactoring in HofX and RTPP that will also be used in the Variational scripts in the next commit

b43fc56 + ddc2a46 + 674124e
Closes #30.
Enables the Variational application to operate on a single member of an ensemble cycling setup OR a standalone single-state cycling setup.  Each member has its own job in a separate run directory under `CyclingDA/DATEDIRECTORY/run`.  Adds an EnsembleOfVariational application that carries out the single EDA executable when `useEnsembleOfVariational` is True, the default choice being False.

0d883bd
Fixed a bug in `config/modeldata.csh`, which only would have been triggered if nEnsDAMembers was set larger than 80.

### Testing
Successfully completed 2 full cycles with the following 4 setups
 - OIE120km
   - DAType: 3denvar
   - DAType: eda_3dvar
     - useEnsembleOfVariational: False
       - (1) nEnsDAMembers: 5
       - (2) nEnsDAMembers: 80
     - useEnsembleOfVariational: True
       - (3) nEnsDAMembers: 5
 - O30kmIE120km
   - (4) DAType: 3denvar


jjguerrette:
Note that the base branch is `feature/variationalPrep`, because this PR depends on #31.

drive.csh
```diff
@@ -359,16 +359,16 @@ cat >! suite.rc << EOF
...
+{% if useEnsembleOfVariational and nEnsDAMembers > 1 %}
   [[CyclingEDA]]
     inherit = CyclingDA
-    script = \$origin/EnsembleVariational.csh
+    script = \$origin/EnsembleOfVariational.csh
     [[[job]]]
-      execution time limit = PT${EnsVariationalJobMinutes}M
-      execution retry delays = ${EnsVariationalRetry}
+      execution time limit = PT${EnsOfVariationalJobMinutes}M
+      execution retry delays = ${EnsOfVariationalRetry}
```
ibanos90, line 368: Got the following error while creating the suite.rc --> "EnsOfVariationalRetry: Undefined variable."
This is because the variable EnsVariationalRetry was changed to EnsOfVariationalRetry in the drive.csh script but this change should also be made in the config/job.csh script. After this change, the suite was created successfully.

jjguerrette: Thanks for the catch.  Fixed it.

ibanos90: Sure! Thank you!

ibanos90 (approved):
Tested the same configurations as in the "Testing" part, but for a few cycles, and all jobs ended successfully!

PR #36, opened 2021-10-18
Enable multiple EDA instances

### LINKED ISSUE #33: Enable sub-ensembles of multiple EDA's
Follow on to #30 for EnsembleOfVariational application

jjguerrette:
With this PR, the EDA members are configured using `EDASize` and `nDAInstances` instead of `nEnsDAMembers`, which is the product of those two new settings.  When `EDASize == 1` (default), an ensemble of independent Variational applications is conducted just as before.  When `EDASize > 1`, multiple instances of the EnsembleOfVariational application are initiated, each with `EDASize` members.

Users now have the option to configure the minimization algorithm in config/experiment.csh as well.  Although one could use `DRPBlockLanczos` (block EDA) for each of the EDA instances, so far that algorithm is not working for our 120km benchmark experiment.  Full testing/investigation remains for future work.

Closes #33

### DIFF
--- a/EnsembleOfVariational.csh
+++ b/EnsembleOfVariational.csh
@@ -1,12 +1,33 @@
 #!/bin/csh -f
 
-# Carry out variational minimization for either
-# multiple first guess states (EDA)
+# Carry out variational minimization for multiple first guess states (EDA)
+# ARGUMENTS:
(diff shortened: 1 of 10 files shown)


ibanos90 (approved):
Tested it with EDASize = 1 and nDAInstances = 20, and, EDASize = 2 and nDAInstances = 10. All changes worked well.

PR #38, opened 2021-10-22
OIE60km EDA and dual-mesh O30kmIE60km single-state 3denvar

jjguerrette:
Enables 60km cycling experiments and dual-mesh cycling experiments with 30km outer loop and forecasts and 60km inner loop and ensembles

Closes #37

PR #40, opened 2021-11-01
Use default ncar_pylib to avoid verification errors

jjguerrette:
Fixes bug reported by @ibanos90 in #38

When `ncar_pylib.csh` or `ncar_pylib` (bash) are sourced, they take any command-line arguments that are given to the parent scripts where the source occurs.

For example,

`script0.csh`:
```
./script1.csh arg1
```

`script1.csh`:
```
source ncar_pylib.csh
```

When `script0.csh` is executed,
```
./script0.csh
```
ncar_pylib.csh will receive the argument `arg1`.  The same is true through multiple layers of `source`.  The arguments to `ncar_pylib` have very specific meanings.  Here the solution is to always give `ncar_pylib` a single argument of `default`, which overrides the arguments given to the parent script.

This issue showed up recently, because CISL removed the previously default behavior of propagating the login environment to job scripts.  So the ncar_pylib environment on a login node no longer transfers to the job environment.  We could instead issue the -V PBS flag to `VerifyObs` and `VerifyModel` job scripts, which I have confirmed also works.  However, it is better to declare the entire environment needed for those tasks instead of depending on the user to set up the environment before submitting the cylc suite.

### DIFF
--- a/GenerateABEInflation.csh
+++ b/GenerateABEInflation.csh
@@ -100,7 +100,7 @@ while ( $success != 0 )
   set success = $?
 
   if ( $success != 0 ) then
-    source /glade/u/apps/ch/opt/usr/bin/npl/ncar_pylib.csh
+    source /glade/u/apps/ch/opt/usr/bin/npl/ncar_pylib.csh default
     sleep 3
   endif
 end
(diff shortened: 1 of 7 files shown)


jjguerrette:
Merging this, because all other PR's depend on the change and it works.

PR #48, opened 2021-11-08
Change all ObsPlugs yaml stubs to be named by instrument alone

### LINKED ISSUE #42: Divide clramsua, cldamsua, and allmhs into individual instrument YAMLS
This will become more important as we add new instruments and want to configure each one individually, especially when adding extensive QC filters to each one.

Will also add some re-usable anchors for all CRTM `obs operator` and all `get values` sections to reduce repetition.  Propagate their used to abi_goes16 and ahi_himawari8.

jjguerrette:
## Description
The original motivation of this PR was to separate all amsua and mhs instruments out to different YAMLS (#42).  During that process, it became clear that ABI and AHI yaml stubs suffered from a related problem, where multiple YAML stubs described a single instrument with the same operator (e.g., all-sky CRTM).  Each stub was associated with a different super-obbing grid size, with the only thing different between files being the all-sky ObsError parameterization.

With this PR, each combination of instrument and operator is described in only one YAML stub, such that users enter the instrument names directly in `config/experiment.csh` instead of, e.g., `cldamsua`, `clramsua`, `all$abi`, and `all$ahi`.  sed substitution is applied to the all-sky ABI and AHI yamls in order to select the correct ObsError parameterization for the experiment-specific combination of super-obbing grid and model mesh.  The super-obbing grid is now automatically selected for each `MPASGridDescriptor` under `config/mpas/$MPASGridDescriptor/mesh.csh`.

When super-obbing is enabled in UFO instead of being a characteristic of the input data, the latest functionality can be modified to simply adjust the super-ob YAML key instead of reading in a unique data set (i.e., the part currently handled in `config/obsdata.csh`).

## Issues
Closes #42

## Testing
Tested 3denvar for OIE120km (2 cycles), OIE60km (only 1 cycle works due to missing SST update files), O30kmIE60km (2 cycles), all while assimilating `abi_g16` and `ahi_himawari8`.  Note that this required modification to some `config/mpas/OIE60km/job.csh` and `config/mpas/O30kmIE60km/job.csh` in order to have enough memory when hydrometeors (`qc`, `qi`, `qg`, `qr`, `qs`) are added as increment variables.  The changes for `O30kmIE60km` are made permanent, but those for `OIE60km` only entered as a commented "note".


ibanos90 (approved):
Run the same tests as in "Testing" and all jobs completed successfully.

PR #49, opened 2021-11-09
Use python script to substitute ensemble B members in YAML

### LINKED ISSUE #20: Separate jediPrep task out of CyclingDA task instead of performing as env-script

jjguerrette:
### Description
Addresses the final item in #20.  Previously expensive cat and sed substitution were used to generate the ensemble B `members` section of the YAML for all EDA members or for a single variational member.  This step acted as a bottleneck before the running EDA Variational applications, because each ensemble member YAML (independent files across members) was handled in serial and the command-line procedures are very slow.  The substitution is moved to a new script, `tools/substituteEnsembleB.py`, which is more robust, easier to read and modify, and MUCH faster.

One added feature not previously feasible with sed substitution is that `substituteEnsembleB` shuffles the order of the ensemble B members in the YAML for each Variational instance, which should reduce the number of simultaneous reads from identical state files.  The shuffling can also be turned off via a command-line argument.

Closes #20

### Testing
Completed a single cycle of DA for the following combinations of settings
- DAType: 3denvar
- DAType: eda_3denvar; nDAInstances: 5; LeaveOneOutEDA: True
- DAType: eda_3denvar; nDAInstances: 5; LeaveOneOutEDA: False
- DAType: eda_3denvar; nDAInstances: 80; LeaveOneOutEDA: True

Performed initial testing with shuffling turned off in `substituteEnsembleB` in order to make sure the correct ensemble B members are selected for different settings of `LeaveOneOutEDA`.  Then turned on shuffling to check that the ensemble B member ordering is different across EDA members.


ibanos90 (approved):
Tested the changes after the last commit and the variable LeaveOneout as False and True is working correctly. Thank you!

PR #54, opened 2021-12-08
Feature/FirstFC_coldstart

### LINKED ISSUE #51: Enable the model forecast to run using cold start initial conditions in addition to warm start
In order to do that the following steps are necessary:
- add a new functionality to the MPAS-Workflow to execute the model based on existing functionalities in the workflow
- verify if changes in the namelist and streams atmosphere are needed and do the modifications as needed
- verify the existence of the initial conditions for the selected data

ibanos90:
This PR enables MPAS-Workflow to create cold start conditions from an external GFS analysis (cold start), in addition to using a pre-generated first forecast (warm start) and to reinitiate the cycling or the suite from any cycle (restart).

**Changes:**
A new variable "InitializationType" is defined in the config/experiment.csh script which can take values ColdStart/WarmStart/ReStart (note that ReStart is meant for the workflow when restarting from a cycle that had previously failed or stopped for any reason). In this PR is ensured that the initial cycle is the same for both initialization types, cold and warm start. In order to do that, several changes were needed in the config/modeldata.csh script in which the first date is used instead of using the previous one. The next date is also introduced.

In addition, part of the SetupWorkflow.csh script was moved to the "warmStart.csh" script, which is the one that creates the link/copy of the pre-generated first forecast. Other changes in SetupWorkflow.csh script were necessary to create the corresponding "CStart.csh" and "WStart.csh" scripts for each initialization type. The forecast.csh script was also modified to allow its use for the first forecast generation.

**Caveat**: 
The namelist.atmosphere for the generation of the first forecast only differs from the one used for warm start in that the variable "config_do_DAcycling" in the &restart block is equal false (otherwise is true). I tried to add this as a variable substitution but it did not work (left the intended changes commented out). Therefore, a new variable "NamelistFileIni" was added in the forecast.csh script and the namelist.atmosphere_Init file is added to account for the config_do_DAcycling = false.

Closes https://github.com/NCAR/MPAS-Workflow/issues/51


ibanos90:
Tests conducted:

Cold start: 3denvar
Warm start: 3denvar and eda_3denvar 
VerifyDeterministicDA = True
Restart: from cycles 20180415T00 and 20180415T06 (after a cold or warm start run) to 20180415T12

config/experiment.csh
```diff
@@ -78,8 +87,8 @@ set l = ($l $benchmarkObsList)
 set variationalObsList = ($l)
 
 ## DAType
-# OPTIONS: 3denvar, eda_3denvar, 3dvarId
-setenv DAType eda_3denvar
+# OPTIONS: 3denvar, eda_3denvar, 3dvarId, 3dvar
+setenv DAType 3denvar
```
jjguerrette: Change back to eda_3denvar.  Remove 3dvar until we actually have it implemented.

config/experiment.csh
```diff
@@ -138,7 +147,7 @@ endif
 
 ## LeaveOneOutEDA, whether to use self-exclusion in the EnVar ensemble B during EDA cycling
 # OPTIONS: True/False
-setenv LeaveOneOutEDA True
+setenv LeaveOneOutEDA False
```
jjguerrette: Change back to True

config/experiment.csh
```diff
@@ -223,7 +232,6 @@ setenv DAVFWindowHR ${CyclingWindowHR}  # window of observations included in AN/
 setenv FCVFWindowHR 6                   # window of observations included in forecast verification
 setenv forecastPrecision single         # floating-point precision of forecast output; options: [single, double]
 
-
```
jjguerrette: Add back line break

drive.csh
```diff
@@ -14,13 +14,12 @@ source config/experiment.csh
...
 # OR:
 # + CyclingFC must have been completed for the cycle before initialCyclePoint. Set > FirstCycleDate to automatically restart#   from a previously completed cycle.
-set initialCyclePoint = 20180415T00
+set initialCyclePoint = 20180414T18
 
 ## finalCyclePoint
 # OPTIONS: >= initialCyclePoint
 # + ancillary model and/or observation data must be available between initialCyclePoint and finalCyclePoint
-set finalCyclePoint = 20180514T18
-
```
jjguerrette: add back line break

drive.csh
```diff
@@ -29,6 +28,7 @@ set finalCyclePoint = 20180514T18
 #                   DA and FC cycling components
 # OPTIONS: Normal, Bypass, Reanalysis, Reforecast
 set CriticalPathType = Normal
+#### another setting for the intialization type
```
jjguerrette: Either clarify or remove comment

drive.csh
```diff
@@ -60,13 +60,13 @@ set CompareBG2Benchmark = False
 ## VerifyEnsMeanBG: whether to run verification scripts for ensemble mean
 #    background (nEnsDAMembers > 1) or deterministic background (nEnsDAMembers == 1)
 # OPTIONS: True/False
-set VerifyEnsMeanBG = True
+set VerifyEnsMeanBG = False
```
jjguerrette: Change back to True

drive.csh
```diff
@@ -60,13 +60,13 @@ set CompareBG2Benchmark = False
...
 
 ## DiagnoseEnsSpreadBG: whether to diagnose the ensemble spread in observation
 #    space while VerifyEnsMeanBG is True.  Automatically triggers HofXBG
 #    for all ensemble members.
 # OPTIONS: True/False
-set DiagnoseEnsSpreadBG = True
+set DiagnoseEnsSpreadBG = False
```
jjguerrette: Change back to True

drive.csh
```diff
@@ -81,12 +81,13 @@ set VerifyExtendedEnsFC = False
 
 date
 
-## Initialize cycling directory if this is the first cycle point
-set yymmdd = `echo ${FirstCycleDate} | cut -c 1-8`
-set hh = `echo ${FirstCycleDate} | cut -c 9-10`
-set firstCyclePoint = ${yymmdd}T${hh}
-if ($initialCyclePoint == $firstCyclePoint) then
-  ./SetupWorkflow.csh
+source SetupWorkflow.csh
```
jjguerrette: Why sourcing instead of executing?  Also, does this need to be executed all the time or only when `$initialCyclePoint == $firstCyclePoint`?

drive.csh
```diff
@@ -81,12 +81,13 @@ set VerifyExtendedEnsFC = False
...
+
+## Set the cycles hours
+if ( ${InitializationType} == "ColdStart" || ${InitializationType} == "WarmStart") then
+  set cyclingCycles = +PT${CyclingWindowHR}H/PT${CyclingWindowHR}H
+else if ( ${InitializationType} == "ReStart" ) then
+  set cyclingCycles = PT${CyclingWindowHR}H
```
jjguerrette: Why do these need to be different?

SetupWorkflow.csh
```diff
@@ -40,44 +41,30 @@ source config/mpas/${MPASGridDescriptor}/mesh.csh
...
+## Generate First FC
+echo "Making First forecast (cold start) job script"
+set JobScript=${mainScriptDir}/CStart.csh
+sed -e 's@WorkDirsTEMPLATE@CyclingFCDirs@' \
+    -e 's@StateDirsTEMPLATE@CyclingDAOutDirs@' \
+    -e 's@fcLengthHRTEMPLATE@'${CyclingWindowHR}'@' \
+    -e 's@fcIntervalHRTEMPLATE@'${CyclingWindowHR}'@' \
+    -e 's@deleteZerothForecastTEMPLATE@True@' \
+    forecast.csh > ${JobScript}
+chmod 744 ${JobScript}
```
jjguerrette: Is there any difference between `CStart.csh` and `CyclingFC.csh`?  If not, then both cylc tasks can use the same csh script.

drive.csh
```diff
@@ -362,6 +368,26 @@ cat >! suite.rc << EOF
       -q = share
       -A = ${VFAccountNumber}
       -l = select=1:ncpus=1
+#Cold Start Initial Forecast
+  [[CStart]]
+    env-script = cd ${mainScriptDir}; ./CStart.csh "1" "clStart"
```
jjguerrette: Following up on my SetupWorkflow.csh comment, would replacing `CStart.csh` with `CyclingFC.csh` on line 373 yield the same result?  If so, there is no need to have two different scripts.

drive.csh
```diff
@@ -362,6 +368,26 @@ cat >! suite.rc << EOF
       -q = share
       -A = ${VFAccountNumber}
       -l = select=1:ncpus=1
+#Cold Start Initial Forecast
+  [[CStart]]
```
jjguerrette: Please change the name of the cylc task from `CStart` to `ColdStartFC`

drive.csh
```diff
@@ -362,6 +368,26 @@ cat >! suite.rc << EOF
...
+    env-script = cd ${mainScriptDir}; ./CStart.csh "1" "clStart"
+    script = \$origin/CStart.csh "1" "clStart"
+    [[[job]]]
+      execution time limit = PT4M
+      execution retry delays = ${StartRetry}
+    [[[directives]]]
+      -m = ae
+      -l = select=1:ncpus=36:mpiprocs=36
+#Warm Start Initial Forecast
+  [[WStart]]
```
jjguerrette: Please change the name of the cylc task and csh script names from `WStart` to `GetWarmStartIC`.  Then the new task that converts GRIB to mpas cold-start init files will be `GenerateColdStartIC`.

forecast.csh
```diff
@@ -70,7 +73,16 @@ cp -v ${memberStaticFieldsFile} ${localStaticFieldsFile}
 set icFileExt = ${fileDate}.nc
 set icFile = ${ICFilePrefix}.${icFileExt}
 rm ./${icFile}
-ln -sfv ${self_icStateDir}/${self_icStatePrefix}.${icFileExt} ./${icFile}
+if ( ${initType} == "clStart" ) then
```
jjguerrette: Please change the `initType` strings as follows: `clStart` to `cold` and `wmStart` to `warm`

forecast.csh
```diff
@@ -103,12 +115,13 @@ sed -i 's@forecastPrecision@'${forecastPrecision}'@' ${StreamsFile}
...
 ## copy/modify dynamic namelist
 rm ${NamelistFile}
-cp -v ${self_ModelConfigDir}/${NamelistFile} .
+cp -v ${self_ModelConfigDir}/${NamelistFileIni} ${NamelistFile}
 sed -i 's@startTime@'${NMLDate}'@' $NamelistFile
 sed -i 's@fcLength@'${config_run_duration}'@' $NamelistFile
 sed -i 's@nCells@'${MPASnCellsOuter}'@' $NamelistFile
 sed -i 's@modelDT@'${MPASTimeStep}'@' $NamelistFile
 sed -i 's@diffusionLengthScale@'${MPASDiffusionLengthScale}'@' $NamelistFile
+#sed -i 's@DAcycling@'${do_DAcycling}'@' $NamelistFile
```
jjguerrette: By default this changes the first instance of `DAcycling` in a line to whatever string is held by `$do_DAcycling`.  There are two instances of `DAcycling` on the same line in the namelist.  So all you need to do is pick a different substitution string, like `configDODACycling` instead of `DAcycling`.

config/modeldata.csh
```diff
@@ -29,19 +29,29 @@ set dd = `echo ${FirstCycleDate} | cut -c 7-8`
 set hh = `echo ${FirstCycleDate} | cut -c 9-10`
 setenv FirstFileDate ${yy}-${mm}-${dd}_${hh}.00.00
 
-## date from which first background is initialized
+## previous date from which first background is initialized
 set prevFirstCycleDate = `$advanceCYMDH ${FirstCycleDate} -${CyclingWindowHR}`
+setenv prevFirstCycleDate ${prevFirstCycleDate}
```
jjguerrette, line 34: Do you mind reminding me where `prevFirstCycleDate` and `prevFirstFileDate` are still used?  I'm wondering if we can remove them.

SetupWorkflow.csh
```diff
@@ -13,6 +13,7 @@ echo ""
 
 rm -rf ${mainScriptDir}
 mkdir -p ${mainScriptDir}
+echo `pwd`
```
jjguerrette: Remove

SetupWorkflow.csh
```diff
@@ -40,44 +41,30 @@ source config/mpas/${MPASGridDescriptor}/mesh.csh
...
-  @ member++
-end
-setenv VARBC_TABLE ${INITIAL_VARBC_TABLE}
 
-#TODO: enable VARBC updating between cycles
-#  setenv VARBC_TABLE ${prevCyclingDADir}/${VarBCAnalysis}
+## Copy/link pre-generated first FC 
+echo "Making First forecast (warm start) job script"
+set JobScript=${mainScriptDir}/WStart.csh
+cp warmStart.csh ${JobScript}
```
jjguerrette: In combination with my comment in drive.csh and previous practice for SetupWorkflow.csh, `warmStart.csh` can simply be named `GetWarmStartIC.csh`.  Then you can remove lines 63-66 and add `GetWarmStartIC.csh` to the `workflowParts` list on line 17.

jjguerrette (changes requested):
Very good work!  Thank you Ivette.

No major revisions needed, though I have lots of small comments about renaming to fit my otherwise non-documented conventions of the repo and a few questions.  FYI, my general approach to naming variables is to make them verbose enough to understand in context.  Sometimes that means avoiding prefixes/suffixes that are already implied by the context, and sometimes that means providing lots of information in the variable, script, or cylc task name (e.g., `GetWarmStartIC`).

jjguerrette:
FYI @ibanos90, you can commit additional changes to this same branch then push them with `git push`.  The PR will update.

SetupWorkflow.csh
ibanos90: Removed

SetupWorkflow.csh
ibanos90: You are right, now both cylc tasks use the same csh script. Thanks for pointing this out!

SetupWorkflow.csh
ibanos90: Sure! Done as recommended.

config/experiment.csh
ibanos90: Done as recommended.

config/experiment.csh
ibanos90: Changed

config/experiment.csh
ibanos90: Added

config/modeldata.csh
ibanos90: Sure! prevFirstCycleDate is still used in the PrepVariational.csh script line 369 for the variable "FirstCyclingFCDir". prevFirstFileDate is only set in config/modeldata.csh and I don't see its use in other scripts. I did a test and it can be removed.

drive.csh
ibanos90: Added

drive.csh
ibanos90: Comment removed

drive.csh
ibanos90: Changed

drive.csh
ibanos90: Changed

drive.csh
ibanos90: You are right, it should be executed. After several attempts, the only way the suite was completely updated for the case of a restart initialization type was when the this script is executed and then all scripts were created again. However, I am not sure why it doesn’t work otherwise. The R1 task for the initial cycle point for the restart remained the same as in the warm start if this is not done.

drive.csh
ibanos90: For a restart initialization type we may want to start just at the cycle that crashed or stopped for any reason, not 6h before. I can certainly think in other solution if you think the first cycle in a restart case should start 6h before. Please let me know what would be the best way to proceed.

drive.csh
ibanos90: Changed

drive.csh
ibanos90: Yes, it does the same and thus I modified it in the drive.csh script.

drive.csh
ibanos90: Changed

forecast.csh
ibanos90: Changed

forecast.csh
ibanos90: You are right. With this change the substitution is successful. Now we can use the same namelist file for both initialization types and just change the configDODACycling variable accordingly. Thank you!

config/modeldata.csh
jjguerrette: Thanks!

ibanos90:
> 

Thanks @jjguerrette! I did a new commit with all the changes you recommended. I think I covered all of them, but please let me know if I left something out. I run the same tests as before and everything is working well.

drive.csh
jjguerrette: You make a good justification.  Do you mind adding comments in the two different branches of the if clause describing that reasoning?  Then we will be reminded how to set the first cycle date and initial cycle point appropriately.

forecast.csh
```diff
@@ -20,6 +20,9 @@ if ( $ArgMember < 1 ) then
   exit 1
 endif
 
+## arg to get the initialization type
```
jjguerrette, line 23: Please add a comment describing valid values for this argument (`cold`, `warm`).  See experiment.csh for similar examples.

jjguerrette:
Thank you @ibanos90 for the changes.  It looks very good, and I added only a few requests for in-line comments.  Looking at the "Files changed" tab on the github page, there is a very small footprint from this PR, which is a great example of how to add new functionality while minimally impacting existing functionality.  I will run some tests too.

drive.csh
ibanos90: Sure! Thanks for that suggestion. I added some comments.

forecast.csh
ibanos90: Added some comments

ibanos90:
Thanks @jjguerrette! I added some in-line comments as suggested in the recent commit.

drive.csh
```diff
@@ -191,21 +195,27 @@ cat >! suite.rc << EOF
...
+{% if InitializationType == "ColdStart" %}
+    [[[R1]]]
+      graph = ColdStartFC => CyclingFCFinished
+{% elif InitializationType == "WarmStart" %}
+    [[[R1]]]
+      graph = GetWarmStartIC => CyclingFCFinished 
+{% elif InitializationType == "ReStart" %}
+    [[[R1]]]
+      graph = '''
+      '''
```
jjguerrette: You have explained this to me multiple times, but I'm still trying to understand why this branch of the if statement is necessary.  It seems like the `cyclingCycles` variable ought to handle all situations without an R1 task for `ReStart`.

drive.csh
```diff
@@ -81,12 +92,16 @@ set VerifyExtendedEnsFC = False
...
 
-## Initialize cycling directory if this is the first cycle point
-set yymmdd = `echo ${FirstCycleDate} | cut -c 1-8`
-set hh = `echo ${FirstCycleDate} | cut -c 9-10`
-set firstCyclePoint = ${yymmdd}T${hh}
-if ($initialCyclePoint == $firstCyclePoint) then
+## Set the cycles hours (cyclingCycles) according to the initialization type defined in config/experiment.csh
+if ( ${InitializationType} == "ColdStart" || ${InitializationType} == "WarmStart") then
+  # For cold or warm start, the cycling starts 6h after the first cycle date (same as the initial cycle point of the suite)
+  # The cycles will run every 6h, starting 6h after the initial cycle point
```
jjguerrette: Maybe replace `6h` with `CyclingWindowHR` or "`CyclingWindowHR` hours"?

drive.csh
```diff
@@ -191,21 +205,27 @@ cat >! suite.rc << EOF
...
+      graph = ColdStartFC => CyclingFCFinished
+{% elif InitializationType == "WarmStart" %}
+    [[[R1]]]
+      graph = GetWarmStartIC => CyclingFCFinished 
+{# elif InitializationType == "ReStart" #}
 #    [[[R1]]]
-#      graph = InitEnsFC => InitCyclingDA => CyclingDA
-#{# endif #}
+#      graph = '''
+#      '''
```
jjguerrette: As discussed in person, delete lines 214-217

getCycleVars.csh
```diff
@@ -62,7 +62,6 @@ while ( $member <= ${nEnsDAMembers} )
 
   set CyclingFCDirs = ($CyclingFCDirs ${CyclingFCDir}${memDir})
   set prevCyclingFCDirs = ($prevCyclingFCDirs ${prevCyclingFCDir}${memDir})
-
```
jjguerrette: Add this line break back

drive.csh
ibanos90: Replaced it with "CyclingWindowHR hours"

drive.csh
ibanos90: Lines deleted

getCycleVars.csh
ibanos90: Added

ibanos90:
Thanks for all the suggestions @jjguerrette! I ran the same tests as before and all seems working ok.

drive.csh
```diff
@@ -14,14 +14,25 @@ source config/experiment.csh
...
 
 
+#########################
+# InitializationType 
+#########################
+# Indicates the type of initialization at the initial cycle: cold, warm, or re- start
+#       cold start: generate first forecast online from an external GFS analysis
+#       warm start: copy a pre-generated forecast
+#          restart: allow to restart the cycling/suite from any cycle 
+#                   (run after a warm start cycle that crashed or stopped for any reason)
```
jjguerrette: Would the following changes be accurate for this comment?
```
#          restart: restart the cycling/suite from any cycle 
#            run from a warm start forecast produced from an already existing workflow, which could
#            have been originally initiated from either a warm or cold start initial condition
```

ibanos90: Yes, this looks much better!

forecast.csh
```diff
@@ -20,6 +20,12 @@ if ( $ArgMember < 1 ) then
...
+#       cold: same as InitializationType = ColdStart in config/experiment.csh
+#       warm: same as InitializationType = WarmStart in config/experiment.csh
```
jjguerrette: These comments should be independent of settings in other scripts.  The descriptions need to be in the context of running a forecast, not in the context of the whole workflow.  For example,
```
#       cold: initial condition is produced from MPAS's init executable (i.e., core_init_atmosphere)
#       warm: initial condition is an output state from a previous MPAS forecast (i.e., core_atmosphere)
```

getCycleVars.csh
```diff
@@ -62,7 +62,7 @@ while ( $member <= ${nEnsDAMembers} )
...
+
```
jjguerrette: No need to add a space at the beginning of the line.

jjguerrette: FYI, these kinds of small formatting changes will show up with `git diff develop`.  It's very useful for eliminating spacing changes.

forecast.csh
ibanos90: Thanks for pointing this out! I used your example of comment.

getCycleVars.csh
ibanos90: Ok. I deleted the space

forecast.csh
jjguerrette: I realize now that my previous advice on writing these comments was vague.

getCycleVars.csh
ibanos90: You are right, I forgot to do the diff with the develop branch before committing the changes. Thanks!

drive.csh
```diff
@@ -14,14 +14,26 @@ source config/experiment.csh
 # + initialCyclePoint must be equal to FirstCycleDate
 # OR:
 # + CyclingFC must have been completed for the cycle before initialCyclePoint. Set > FirstCycleDate to automatically restart#   from a previously completed cycle.
```
jjguerrette: This old comment is not exactly accurate any more.  I will pull this branch to test and also fix this comment.

ibanos90: That's right. Thank you!

forecast.csh
```diff
@@ -70,7 +76,14 @@ cp -v ${memberStaticFieldsFile} ${localStaticFieldsFile}
 set icFileExt = ${fileDate}.nc
 set icFile = ${ICFilePrefix}.${icFileExt}
 rm ./${icFile}
-ln -sfv ${self_icStateDir}/${self_icStatePrefix}.${icFileExt} ./${icFile}
+if ( ${initType} == "cold" ) then
+  set initialState = ${localStaticFieldsFile}
```
jjguerrette, line 80: I see that in your next feature branch this `localStaticFieldsFile` usage gets replaced.  So I will wait to see where you go in the next steps.  That was my final comment.  Running tests now.

ibanos90: Yes, it will be replaced by the cold start initial condition created with mpas_init. Thank you for all of your comments!

jjguerrette:
I made a few small nuanced changes.  I tested the 3denvar cold-start, 3denvar warm-start, and eda_3denvar warm-start for 3 cycles.  I also restarted from the second cycle of the 3denvar cold- and warm-starts.

jjguerrette (approved):
Thank you for the hard work!  Feel free to check over my last commit.  Merge when you are ready.

ibanos90:
Thanks for the changes in the last commit @jjguerrette!

PR #55, opened 2021-12-15
Feature/MPAS_init_coldstart

### LINKED ISSUE #52: Generate initial conditions using mpas_init
Add a new functionality to the MPAS-Workflow to generate initial conditions using mpas_init. Here the namelist and streams atmosphere also need to be checked. This is related to https://github.com/NCAR/MPAS-Workflow/issues/51#issue-1058783790.

ibanos90:
With this PR, a new functionality is added to the MPAS-Workflow to generate cold start initial conditions using mpas_init. At this moment, pre-ungribed GFS analysis are used (in a following PR that step is part of the workflow).

**Changes:** 
- A new folder is created in which the pre-ungribed GFS analysis is linked and the mpas_init is executed (this can be changed in the future as needed). For that, changes were needed in the scripts config/filestructure.csh, getCycleVars.csh and config/filestructure.csh;
- The GenerateColdStartIC.csh script that manages the generation of the cold initial conditions was created and added to the workflow parts in SetupWorkflow.csh and the drive.csh accordingly;
- The corresponding namelist.init_atmosphere and streams.init_atmosphere were added and adjusted.

**Tests:**
Tested the 3denvar for ColdStart, WarmStart, and ReStart, and eda_3denvar WarmStart and ReStart for 4 and 3 cycles, respectively.

Closes https://github.com/NCAR/MPAS-Workflow/issues/52


GenerateColdStartIC.csh
```diff
@@ -0,0 +1,73 @@
...
+# Setup environment
+# =================
+source config/experiment.csh
+source config/filestructure.csh
+source config/tools.csh
+source config/modeldata.csh
+source config/mpas/variables.csh
+source config/mpas/${MPASGridDescriptor}/mesh.csh
+source config/builds.csh
+source config/environment.csh
```
jjguerrette: Are all of these needed?  I know it is tedious to figure out, but it is preferable to use a minimal list.

GenerateColdStartIC.csh
```diff
@@ -0,0 +1,73 @@
...
+set yymmdd = `echo ${CYLC_TASK_CYCLE_POINT} | cut -c 1-8`
+set hh = `echo ${CYLC_TASK_CYCLE_POINT} | cut -c 10-11`
+set thisCycleDate = ${yymmdd}${hh}
+set thisValidDate = ${thisCycleDate}
+source ./getCycleVars.csh
+
+# templated work directory
```
jjguerrette: In my informal tech-speak-jargon, this is a "static" work directory since no sed substitution is used to replace a "TEMPLATE" variable that would make this a "templated" work directory.

drive.csh
```diff
@@ -24,7 +24,9 @@ if ( ${InitializationType} == "ColdStart" || ${InitializationType} == "WarmStart
   ./SetupWorkflow.csh
 
   # The initialCyclePoint is the same as FirstCycleDate when starting a new experiment
-  set initialCyclePoint = ${FirstCycleDate}
+  set yymmdd = `echo ${FirstCycleDate} | cut -c 1-8`
+  set hh = `echo ${FirstCycleDate} | cut -c 9-10`
+  set initialCyclePoint = ${yymmdd}T${hh}
```
jjguerrette, line 29: Thank you for fixing this.  Sorry for breaking it.

getCycleVars.csh
```diff
@@ -9,6 +9,7 @@ setenv prevCycleDate ${prevCycleDate}
 #setenv nextCycleDate ${nextCycleDate}
 
 ## setup cycle directory names
+set InitICDir = (${InitICWorkDir}/${thisCycleDate})
```
jjguerrette, line 12: I think you talked about changing `InitICDir` to `GFSAnalysesDir` or something like that in a future PR.  Just making a note for my own reference.  Looks good as-is for this PR.

getCycleVars.csh
```diff
@@ -57,6 +58,7 @@ set BenchmarkVerifyEnsFCDirs = ()
 set member = 1
 while ( $member <= ${nEnsDAMembers} )
   set memDir = `${memberDir} $DAType $member`
+
```
jjguerrette: Remove line break.

jjguerrette (commented):
Mostly looks ready.  A couple comments about the includes/environment section of `GenerateColdStartIC.csh`.

jjguerrette:
As we only expect this PR to affect ColdStart workflows, do you mind adding a description in the PR message of cold-start tests that you have exercised?

ibanos90:
> 

Sure! I forgot to include that in the PR comment. Thanks for reminding me!

GenerateColdStartIC.csh
ibanos90: You are right. I removed sourcing config/tools.csh and config/mpas/variables.csh as they are not really needed in this script

GenerateColdStartIC.csh
ibanos90: I think I got it right now. Please let me know if it looks ok after the last commit

drive.csh
ibanos90: Sure! It's fine!

getCycleVars.csh
ibanos90: At this point, we are creating this InitICDir directory where the cold initial conditions for the first cold start are created. We can certainly modify this in the future if we want to create the initial conditions offline for a different period.

getCycleVars.csh
ibanos90: Removed

ibanos90:
> > 
> 
> Sure! I forgot to include that in the PR comment. Thanks for reminding me!

Added a **Test** section in the PR comment.

drive.csh
```diff
@@ -469,6 +471,11 @@ cat >! suite.rc << EOF
     inherit = CleanBase
     script = \$origin/CleanVariational.csh
   # forecast-related components
+  [[GenerateColdStartIC]]
+    inherit = CyclingFCBase
```
jjguerrette: I would expect GenerateColdStartIC to run faster than CyclingFCBase-type tasks, because there is no model forecast.  The difference in job duration will be more significant for finer meshes.

Therefore, I think this task can not inherit from CyclingFCBase and that all `config/mpas/O*IE*km/job.csh`'s need new entries for estimated durations for this kind of task.  You may start by adding a relevant time and number of processors/nodes to `config/mpas/OIE120km/job.csh`, then just copying it for all others.  Then this cylc task would look something like
```
  [[GenerateColdStartIC]]
    script = \$origin/GenerateColdStartIC.csh
    [[[job]]]
      execution time limit = PT${InitFCJobMinutes}M
      execution retry delays = ${InitFCRetry}
    [[[directives]]]
      -m = ae
      -l = select=${InitFCNodes}:ncpus=${InitFCPEPerNode}:mpiprocs=${InitFCPEPerNode}
```

ibanos90: You are right. I will modify that

drive.csh
```diff
@@ -16,15 +16,17 @@ source config/experiment.csh
 #     ReStart - restart the cycling/suite from any cycle
 #               run from a warm start forecast produced within an already existing workflow, which
 #               was originally initiated from either a warm or cold start initial condition
-set InitializationType = WarmStart
+set InitializationType = ColdStart
```
jjguerrette: Changes look good.  After your testing, this needs to be reverted to WarmStart for now.  Maybe we will make ColdStart the default when the real-time workflow is fully enabled.

ibanos90: That makes sense. I forgot to change that option back, will do it now. Thanks!

jjguerrette (approved):

PR #59, opened 2022-01-04
Use WPS ungrib to decode GFS analysis on RDA

### LINKED ISSUE #53: Use WPS ungrib to decode the GFS analysis
Add a new functionality to the MPAS-Workflow to use WPS ungrib to decode the GFS analysis in grib format which will be used to cold start the MPAS model forecast. This task will search for the availability of GFS analysis on RDA archives on Cheyenne. This is related to https://github.com/NCAR/MPAS-Workflow/issues/51#issue-1058783790 and https://github.com/NCAR/MPAS-Workflow/issues/52#issue-1058785460

ibanos90:
This PR enables the use of WPS ungrib to decode the GFS analysis in grib format to generate the cold start initial conditions for the MPAS model forecast. GFS analysis on RDA archives on Cheyenne are used instead of pre-downloaded and ungribbed GFS analyses. This allows to run experiments for any date available on RDA archive. Nevertheless, this is still only working for the data period of the pre-generated data since we are using 3denvar and it needs the ensemble forecasts for the covariance matrix. With the implementation of 3dvar in the future, we can then run experiments for any date.

**Changes**:
- The UngribColdStartIC.csh script was created to ungrib the GFS analysis using the ungrib executable. The corresponding namelist was added to the folder config/mpas/init;
- The static data files specification in config/modeldata.csh was modified in order to specify the data source according to the initialization type;
- The if statement for the initialization type in the forecast.csh script was modified in order to specify the correct initial state when using cold start for the first date or any other date. The option to update the sea ice was added in this statement (this may change with the surfaces generation)

**Caveats**:
For now, the sea ice is not updated in this application. It will be implemented in the near future.

Closes https://github.com/NCAR/MPAS-Workflow/issues/53


getCycleVars.csh
```diff
@@ -93,3 +94,9 @@ set fileDate = ${yy}-${mm}-${dd}_${hh}.00.00
 set NMLDate = ${yy}-${mm}-${dd}_${hh}:00:00
 set ConfDate = ${yy}-${mm}-${dd}T${hh}:00:00Z
 set ICfileDate = ${yy}-${mm}-${dd}_${hh}
+
+if ( ${InitializationType} == "ColdStart" ) then
+  setenv StaticFieldsDirOuter ${GFSAnaDirOuter}/${FirstCycleDate}
+  setenv StaticFieldsDirInner ${GFSAnaDirInner}/${FirstCycleDate}
+endif
```
jjguerrette: Since the same value persists for all cycles, `StaticFieldsDirOuter` and `StaticFieldsDirInner` can be set in `config/modeldata.csh` when `${InitializationType} == "ColdStart"`.  `getCycleVars.csh` is used for setting variables that differ in each cycle.

`GFSAnaDirVerify` on line 102 needs to use `GFSAnaDirOuter` instead of `GFSAnaDirEnsemble`.  Additionally, the only place that `GFSAnaDirVerify` is used is in `verifymodel.csh`.  `thisValidDate` is used therein to determine the appropriate date.  So I think line 102 may need to be removed OR modified then moved to `config/modeldata.csh`.

forecast.csh
```diff
@@ -62,26 +56,35 @@ set self_ModelConfigDir = $forecastModelConfigDir
...
+  ln -sfv ${initialState} ${localStaticFieldsFile}
+  set updateSea = 0
+else if ( ${InitializationType} == "ColdStart" & ${thisValidDate} != ${FirstCycleDate}) then
+  set initialState = ${self_icStateDir}/${self_icStatePrefix}.${icFileExt}
+  set do_DAcycling = "true"
+  set StaticMemDir = `${memberDir} ensemble $ArgMember "${staticMemFmt}"`
+  set memberStaticFieldsFile = ${StaticFieldsDirOuter}${StaticMemDir}/${StaticFieldsFileOuter}
+  ln -sfv ${memberStaticFieldsFile} ${localStaticFieldsFile}${OrigFileSuffix}
+  cp -v ${memberStaticFieldsFile} ${localStaticFieldsFile}
+  set updateSea = 0
```
jjguerrette: This all looks good for now.  Do you expect that the only change will be to set `updateSea` to 1 on line 79 once that functionality is enabled?

config/ReStartModelData.csh
```diff
@@ -0,0 +1 @@
+WarmStartModelData.csh
```
jjguerrette: This will be a problem if we ever wish to restart a workflow initiated from a ColdStart forecast. I think the only two relevant values of `InitializationType` are `ColdStart` and `WarmStart`.  It can be determined if this is a restart simply by comparing the `initialCyclePoint` and `FirstCycleDate`.  That affects the R1 task...probably something like below.  You had a blank `else` clause there before, but I asked you to delete it.  My mistake.  Except I do not think there needs to be a `ReStart` `InitializationType`.

```
{% if initialCyclePoint == FirstCycleDate %} #might need to modify to get both dates in the same format
  {% if InitializationType == "ColdStart" %}
      [[[R1]]]
        graph = GenerateColdStartIC => ColdStartFC => CyclingFCFinished
  {% elif InitializationType == "WarmStart" %}
      [[[R1]]]
        graph = GetWarmStartIC => CyclingFCFinished
  {% endif %}
{% endif %}
```

getCycleVars.csh
```diff
@@ -2,6 +2,7 @@
 
 source config/filestructure.csh
 source config/tools.csh
+source config/${InitializationType}ModelData.csh
```
jjguerrette: If this is actually required, probably `config/experiment.csh` ought to be source first.  Maybe with the changes suggested below the `*ModelData` will not be needed.

jjguerrette (changes requested):
Pretty close to ready.  Only a few minor changes.

getCycleVars.csh
ibanos90: You are right. I moved the if statement when ${InitializationType} == "ColdStart" to config/modeldata.csh. In addition, I moved all the data source specifications back to config/modeldata.csh. Please let me know if it looks good like this.

In the case of GFSAnaDirVerify, I changed it to use GFSAnaDirOuter for WarmStart but kept it as GFSAnaDirEnsemble for ColdStart in order to use the GFS analyses already in ${EnsembleModelData}. In the future we can move this variable to verifymodel.csh in order to use the valid date

forecast.csh
ibanos90: So far I think that would be the main change

config/ReStartModelData.csh
ibanos90: I agree that we should only have ColdStart and WarmStart initialization types, thus I removed the ReStart type. Following your suggestion I modified drive.csh and now the R1 tasks and cylc cycling cycles depend on the dates

getCycleVars.csh
ibanos90: After moving all the data source specifications back to config/modeldata.csh, sourcing this file is in fact not longer needed

config/modeldata.csh
```diff
@@ -5,22 +5,27 @@ source config/filestructure.csh
...
-
-set GFSAnaDirOuter = ${OuterModelData}/GFSAna
-set GFSAnaDirInner = ${InnerModelData}/GFSAna
 set GFSAnaDirEnsemble = ${EnsembleModelData}/GFSAna
+if ( ${InitializationType} == "ColdStart" ) then
+  set OuterModelData = ${ExpDir}/${MPASGridDescriptorOuter}
+  set InnerModelData = ${ExpDir}/${MPASGridDescriptorInner}
+  set GFSAnaDirOuter = ${OuterModelData}/GFSAna
+  set GFSAnaDirInner = ${InnerModelData}/GFSAna
+  setenv GFSAnaDirVerify ${GFSAnaDirEnsemble}
```
jjguerrette, line 22: I am under the impression that `GFSAnaDirVerify` should always be equal to `GFSAnaDirOuter` in fully functioning WarmStart and ColdStart situations.  Is this a holdover until the ColdStart automatically generates the `GFSAnaDirOuter` data on the fly?  If so, please add a comment to that effect.  If I am mistaken, do you mind explaining?

forecast.csh
jjguerrette: Please add a comment ~~on this line~~ between lines 78/79, something like the following.

```
  TODO(IHB): enable sea surface updating for ColdStart and set updateSea to 1 below
```

This is a useful indicator for the reviewer to know that some work will be done in a follow-up PR.

ibanos90: This is a good point. I added the suggested line to this script

config/modeldata.csh
ibanos90: Yes, you are right. This is holdover until the GFS analyses from RDA are ungribbed and mpas_init executed each cycle for verification purposes. Then, GFSAnaDirVerify will be equal to GFSAnaDirOuter for both initialization types. As suggested, I added a comment to the script

config/builds.csh
```diff
@@ -15,6 +15,10 @@ setenv BuildCompiler 'gnu-openmpi'
 
 set commonBuild = /glade/work/guerrett/pandac/build/mpas-bundle_gnu-openmpi_09JAN2022
 
+# Ungrib
+setenv ungribEXE ungrib.exe
+setenv ungribBuildDir /glade/work/guerrett/pandac/data/GEFS
```
jjguerrette: What do you think about using the ungrib.exe and link_ungrib.csh from the directory below?

```
/glade/u/home/wrfhelp/PRE_COMPILED_CODE/WPSV4.0_intel_serial_large-file/
```

config/modeldata.csh
```diff
@@ -42,8 +45,15 @@ set Ndd = `echo ${nextFirstCycleDate} | cut -c 7-8`
...
-setenv ungribDir /glade/p/mmm/parc/liuz/pandac_common/ungrib_GFS_o3
+## RDA data on Cheyenne
+setenv RDAdataDir /gpfs/fs1/collections/rda/data
+
+## RDA GFS forecasts
+setenv GFSgribdirRDA ${RDAdataDir}/ds084.1
+
+## linkWPS and Vtable files paths
+setenv VtableDir /glade/u/home/schwartz/MPAS_scripts
+setenv linkWPSdir /glade/scratch/mpasrt/wps
```
jjguerrette: Since link_grib.csh is an executable script that is packaged with the WPS source code, probably good to co-locate the `linkWPSdir` with `ungribBuildDir`.  Maybe even rename both of them to WPSBuildDir.

drive.csh
```diff
@@ -5,43 +5,16 @@
...
-endif
+######################
+# workflow date bounds
+######################
+## initialCyclePoint
+# OPTIONS: >= FirstCycleDate (see config/experiment.csh)
+# Either:
+# + initialCyclePoint must be equal to FirstCycleDate
+# OR:
+# + CyclingFC must have been completed for the cycle before initialCyclePoint. Set > FirstCycleDate to automatically restart#   from a previously completed cycle.
```
jjguerrette: Re-insert line break before
```
#   from a previously completed cycle.
```

drive.csh
```diff
@@ -108,6 +81,23 @@ set VerifyExtendedEnsFC = False
...
+  # Create the experiment directory and cylc task script
+  ./SetupWorkflow.csh
+  # The cycles will run every CyclingWindowHR hours, starting CyclingWindowHR hours after the
+  # initialCyclePoint
+  set cyclingCycles = +PT${CyclingWindowHR}H/PT${CyclingWindowHR}H
+else
+  # The cycles will run every CyclingWindowHR hours, starting at the initialCyclePoint
+  set cyclingCycles = PT${CyclingWindowHR}H
+endif
+
```
jjguerrette, line 101: Looks good.  Nice and simple.

config/builds.csh
ibanos90: Sure! I will test it to make sure it works well

config/modeldata.csh
ibanos90: That makes sense. I will make these changes and commit the changes

forecast.csh
```diff
@@ -62,26 +56,36 @@ set self_ModelConfigDir = $forecastModelConfigDir
...
 set icFile = ${ICFilePrefix}.${icFileExt}
 rm ./${icFile}
-if ( ${initType} == "cold" ) then
+if ( ${InitializationType} == "ColdStart" & ${thisValidDate} == ${FirstCycleDate}) then
   set initialState = ${InitICDir}/${InitFilePrefixOuter}.${icFileExt}
   set do_DAcycling = "false"
-else if ( ${initType} == "warm" ) then
+  ln -sfv ${initialState} ${localStaticFieldsFile}
+  set updateSea = 0
+else if ( ${InitializationType} == "ColdStart" & ${thisValidDate} != ${FirstCycleDate}) then
```
jjguerrette: I'm not totally sure about the difference between bitwise and logical and (`&` vs. `&&` according to https://www2.cs.duke.edu/csl/docs/unix_course/intro-130.html), but usually I use the latter.

So maybe change `&` to `&&`.

config/builds.csh
ibanos90: The executable from that directory gives the following error:
`./ungrib.exe: error while loading shared libraries: libimf.so: cannot open shared object file: No such file or directory`
I remember I got this error before while trying to use a different executable than the one in /glade/work/guerrett/pandac/data/GEFS

jjguerrette: OK, we can use the one I built for now.  Probably we need to load specific modules for th PRE_COMPILED_CODE.  Thanks for trying.

config/modeldata.csh
jjguerrette: Based on the PRE_COMPILED_CODE, it is OK for the two directories to be different for now.  Perhaps when you move `linkWPSdir` to config/builds.csh, could you add a TODO to remind us to eventually get those from the same build?

ibanos90: There is a link_grib.csh script in /glade/work/guerrett/pandac/data/GEFS which I can use for now. Then, the same directory can be used for both the link_grib.csh script and the ungrib executable.

drive.csh
ibanos90: Line break re-inserted

forecast.csh
ibanos90: Changed & to && in lines 67 and 72

getCycleVars.csh
```diff
@@ -2,13 +2,15 @@
...
 source config/tools.csh
+source config/modeldata.csh
 
 set prevCycleDate = `$advanceCYMDH ${thisCycleDate} -${CyclingWindowHR}`
 #set nextCycleDate = `$advanceCYMDH ${thisCycleDate} ${CyclingWindowHR}`
 setenv prevCycleDate ${prevCycleDate}
 #setenv nextCycleDate ${nextCycleDate}
 
 ## setup cycle directory names
+set InitICDir = (${InitICWorkDir}/${thisCycleDate})
```
jjguerrette: Why is `InitICDir` moved back to getCycleVars.csh?  It is only used in forecast.csh when `${thisValidDate} == ${FirstCycleDate}`.  So there will always be a specific date associated with it.

ibanos90: I needed to include the date to InitICDir because in config/modeldata.csh it was equal to only ${GFSAnaDirOuter}.
I changed it as it was before but I am adding `setenv InitICDir ${InitICWorkDir}/${FirstCycleDate}` in config/modeldata.csh in order to create the directory with the date associated for the initial conditions. In the future it needs to be modified in order to create the initial conditions for each cycle. Does that makes sense?

jjguerrette: It makes sense that `${GFSAnaDirOuter}/${thisCycleDate}` would be used at multiple cycles to create the initial conditions at each cycle for verification.  Moving "`${InitICWorkDir}/${FirstCycleDate}` in `config/modeldata.csh`" as you propose in the above comment is all I was suggesting.  Again, if there are steps to be completed in a follow-on PR, it's helpful to make note of them with a TODO.

forecast.csh
```diff
@@ -62,26 +56,36 @@ set self_ModelConfigDir = $forecastModelConfigDir
...
+  set updateSea = 0
+else if ( ${InitializationType} == "WarmStart" ) then
   set initialState = ${self_icStateDir}/${self_icStatePrefix}.${icFileExt}
   set do_DAcycling = "true"
+  set StaticMemDir = `${memberDir} ensemble $ArgMember "${staticMemFmt}"`
+  set memberStaticFieldsFile = ${StaticFieldsDirOuter}${StaticMemDir}/${StaticFieldsFileOuter}
+  ln -sfv ${memberStaticFieldsFile} ${localStaticFieldsFile}${OrigFileSuffix}
+  cp -v ${memberStaticFieldsFile} ${localStaticFieldsFile}
+  set updateSea = 1
 endif
```
jjguerrette: It looks like the only difference between these two `else if` branches is the `updateSea` value.  Other code is the same and duplication could be removed.

jjguerrette (commented):
One final comment.  Feel free to address in a follow-on PR.  Great work!

After the passing test list is added to the PR description, I'm ready to approve.  Let me know if I can help run some tests.

Hopefully cold/warm and first/restart dates have been tested for 120km.

getCycleVars.csh
ibanos90: Thanks! I added a comment

forecast.csh
ibanos90: You are right. I made some changes and moved updateSea back to config/modeldata.csh

jjguerrette: Thanks

ibanos90:
> 

I added the list of tests conducted to the PR description. Thanks for your review!

jjguerrette (approved):
Thanks you.  Before merging, please update the title to something related to the work being completed by this PR.

PR #61, opened 2022-01-11
New build to avoid errors related to libpnetcdf.so.4

### LINKED ISSUE #60: BUG: Adjust TemplateFieldsFileInner cycle date in PrepVariational
https://github.com/NCAR/MPAS-Workflow/blob/35be181d85eea73038180a12ebf715d3bee35fb9/PrepVariational.csh#L369

In order to fix the dual-mesh DA application (i.e. O30kmIE60km), the following change needs to be made starting from the line referenced above.

```
-    set FirstCyclingFCDir = ${CyclingFCWorkDir}/${prevFirstCycleDate}${memDir}/Inner
-    cp -v ${FirstCyclingFCDir}/${self_StatePrefix}.${FirstFileDate}.nc $tFile
+    set FirstCyclingFCDir = ${CyclingFCWorkDir}/${FirstCycleDate}${memDir}/Inner
+    cp -v ${FirstCyclingFCDir}/${self_StatePrefix}.${nextFirstFileDate}.nc $tFile
```

After that change, there may no longer be a need for variables related to `prevFirstCycleDate`.  If true, then `prevFirstCycleDate` and all associated commands can be removed from `config/modeldata.csh`.

jjguerrette:
Over the weekend errors started showing up that look like this:
```
./mpasjedi_hofx3d.x: error while loading shared libraries: libpnetcdf.so.4: cannot open shared object file: No such file or directory
--------------------------------------------------------------------------
Primary job  terminated normally, but 1 process returned
a non-zero exit code. Per user-direction, the job has been aborted.
--------------------------------------------------------------------------
```

The same errors occurred in the bundle build directory when running ctests.  Rebuilding that old directory or making an entirely new build solve the problem, although I do not know why.

List of changes:
- This PR uses a new build from 09JAN2022.
- `config/mpas/geovars.yaml` is updated to be consistent with `mpas-jedi:develop`.
- Fixes #60

### DIFF
--- a/GetWarmStartIC.csh
+++ b/GetWarmStartIC.csh
@@ -30,7 +30,7 @@ while ( $member <= ${nEnsDAMembers} )
         set innerFCDir = $CyclingFCDirs[$member]/Inner
         mkdir -p ${innerFCDir}
         set fcFile = $innerFCDir/${FCFilePrefix}.${nextFirstFileDate}.nc
-        set InitialMemberFC = "$firstFCDirOuter"`${memberDir} ens $member "${firstFCMemFmt}"`
+        set InitialMemberFC = "$firstFCDirInner"`${memberDir} ens $member "${firstFCMemFmt}"`
         ln -sfv ${InitialMemberFC}/${firstFCFilePrefix}.${nextFirstFileDate}.nc ${fcFile}${OrigFileSuffix}
         # rm ${fcFile}
         cp ${fcFile}${OrigFileSuffix} ${fcFile}
--- a/PrepVariational.csh
+++ b/PrepVariational.csh
@@ -366,8 +366,8 @@ while ( $member <= ${nEnsDAMembers} )
     #modify "Inner" initial forecast file
     # TODO: capture the naming convention for FirstCyclingFCDir somewhere else
     set memDir = `${memberDir} $DAType 1`
-    set FirstCyclingFCDir = ${CyclingFCWorkDir}/${prevFirstCycleDate}${memDir}/Inner
-    cp -v ${FirstCyclingFCDir}/${self_StatePrefix}.${FirstFileDate}.nc $tFile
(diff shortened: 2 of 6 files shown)


ibanos90 (approved):
Tested it for OIE120km and O30kmIE60km with 3denvar (2 cycles) and all tasks completed successfully. Just to document it, I also tried O30kmIE120km, but the forecast failed during the second cycle.

jjguerrette:
> Tested it for OIE120km and O30kmIE60km with 3denvar (2 cycles) and all tasks completed successfully. Just to document it, I also tried O30kmIE120km, but the forecast failed during the second cycle.

Thank you for your thorough checks!

ibanos90:
> 

Sure!

jamiebresch:
I just ran into this `error while loading shared libraries: libpnetcdf.so.4: cannot open shared object file: No such file or directory` when I tried to run the mpas-jedi code I built in Sept 2021.
After loading the `module load jedi/gnu-openmpi/9.1.0` used for building the code, `module swap pnetcdf/1.12.2 pnetcdf/1.12.1` is needed to load the consistent pnetcdf/1.12.1  that was loaded as part of jedi/gnu-openmpi/9.1.0 for building the code.

jjguerrette:
> I just ran into this `error while loading shared libraries: libpnetcdf.so.4: cannot open shared object file: No such file or directory` when I tried to run the mpas-jedi code I built in Sept 2021. After loading the `module load jedi/gnu-openmpi/9.1.0` used for building the code, `module swap pnetcdf/1.12.2 pnetcdf/1.12.1` is needed to load the consistent pnetcdf/1.12.1 that was loaded as part of jedi/gnu-openmpi/9.1.0 for building the code.

Thank you for figuring that out and it is good to have a data point on the ever-changing modules on Cheyenne.  It is something we will need to consider when deciding how to support older builds in the workflow.

PR #62, opened 2022-01-27
Enable 3DVAR DA

ibanos90:
This PR enables the 3dvar application in the workflow. The latest static background statistic files are used.

**Changes**:
- The background error section was modified to include the static B statistics in config/applicationBase/3dvarId.yaml
 
**Tests**: 
- 3dvar WarmStart and ColdStar for various cycles


jjguerrette:
`3dvarId` was a reference to the identity B, which is not an option we need to keep around.

Do you mind running the following?

`git mv config/applicationBase/3dvarId.yaml config/applicationBase/3dvar.yaml`

Then also please change `3dvarId` to `3dvar` in the list of `DAType` options in `config/experiment.csh`, with a comment that `3dvar` only works for OIE120km so far.

jjguerrette:
Another thought...

What do you think about making `3dvar` the default option for `DAType`?  It is the least cumbersome option for somebody to start with a new experiment period.

ibanos90:
> 

Sure, that makes sense. I will leave it as the default option.

ibanos90:
> 

Ok, I will commit those changes.

jjguerrette (approved):
Looks good.  Thank you.

PR #63, opened 2022-02-04
Add observation preprocessing from RDA

### LINKED ISSUE #56: Preprocessing observations from RDA
Add a new functionality to the MPAS-Workflow to preprocess RDA BUFR observations to IODA-v1 format  using Jamie's obs2ioda-v1 converter (https://github.com/jamiebresch/obs2ioda/tree/main/obs2ioda-v1).

ibanos90:
**Description**:
This PR adds the preprocessing of RDA archived NCEP BUFR observations to IODA-v2 format and their assimilation into the workflow. Jamie's obs2ioda converter (https://github.com/jamiebresch/obs2ioda) is used to convert prepbufr, satwnd, satwind, and 1bamua (default observations in the list provided for now, others can be uncommented and converted as well) observations and then the ioda-upgrade.x executable is used to get the correct string station_id and string variable_names of the preburf observations. This allows to run experiments for any date using the ColdStart initialization type and it can also be used for WarmStart cases. The preprocessing of the observations listed as default takes around 2 minutes of walltime. The walltime increases as more observations are added, but still is less than 10 minutes even with IASI observations.

**Changes**:
- A new script based on rda_obs2ioda.csh (Jamie's script) was added to convert the observations to IODA-v2 and the folder with the new observations is located in the experiment run directory. 
- Adjustments were made in PrepJEDI.csh to account for the proper data source, taking into account if the option to preprocess the observations is on/off.

**Test**:
- Various tests were conducted:
 Cold Start and the new converted observations for different dates ==> OIE120km, 3dvar 
 WarmStart using pre-generated observations and the new converted observations ==> OIE120km, 3dvar, 3denvar
 
**Caveats**:
- The model space verification still is not working for other dates because of the need to generate the GFS analysis for each cycle for the new dates. This will be the subject of future developments.

Closes https://github.com/NCAR/MPAS-Workflow/issues/56


ObstoIODA.csh
```diff
@@ -0,0 +1,188 @@
...
+# OBTYPES: observation types to pre-process
+set OBTYPES = "$1"
+
+# Setup environment
+# =================
+source config/experiment.csh
+source config/filestructure.csh
+source config/tools.csh
+source config/modeldata.csh
+#source config/obsdata.csh
```
jjguerrette: This commented line can be removed.

ObstoIODA.csh
```diff
@@ -0,0 +1,188 @@
...
+## args
+# OBTYPES: observation types to pre-process
+set OBTYPES = "$1"
+
+# Setup environment
+# =================
+source config/experiment.csh
+source config/filestructure.csh
+source config/tools.csh
+source config/modeldata.csh
```
jjguerrette, line 11: What is needed from modeldata.csh?

config/ObsPlugs/hofx/sfc.yaml
```diff
@@ -0,0 +1,21 @@
+- obs space:
```
jjguerrette: If sfc.yaml replaces sfcp.yaml, then usually I would use a `git mv` to change the file name and maintain the git history of the file.  In any case, the `sfcp.yaml` files can be removed _if_ they are now obsolete.  Same for the variational version.

FYI, I just looked in the diagnostic python scripts.  We already use `sfc` instead of `sfcp` there.  So surface pressure diagnostics should work with this change.

ObstoIODA.csh
```diff
@@ -0,0 +1,188 @@
...
+
+# write out hourly files for IASI
+setenv OPTIONS "-split"
+
+foreach inst ( ${OBTYPES} )
+
+  if ( ${inst} == satwnd ) then
+     setenv THIS_FILE gdas.${inst}.t${hh}z.${ccyy}${mmdd}.bufr
+     if ( ! -e ${THIS_FILE}) then
+        echo "Source file: ${RDAdataDir}/ds351.0/bufr/${ccyy}/${THIS_FILE}"
```
jjguerrette, line 39: Where is `RDAdataDir` defined?

config/experiment.csh
```diff
@@ -51,14 +51,26 @@ set l = ($l sondes)
...
+set l = ()
+set l = ($l prepbufr)
+set l = ($l satwnd)
+set l = ($l satwind)
+set l = ($l 1bamua)
+#set l = ($l 1bmhs)
+#set l = ($l airsev)
+#set l = ($l cris)
+#set l = ($l mtiasi)
+set obtypes = ($l)
```
jjguerrette: Do you mind using a more informative name than `obtypes`?  Maybe `preprocessObsList` or something similar?

config/mpas/O30kmIE120km/job.csh
```diff
@@ -6,10 +6,15 @@ source config/modeldata.csh
...
 
-@ InitICJobMinutes = 1
+setenv InitICJobMinutes 1
 setenv InitICNodes 1
 setenv InitICPEPerNode 36
 
+setenv ObstoIODAJobMinutes 10
+setenv ObstoIODANodes 1
+setenv ObstoIODAPEPerNode 1
+setenv ObstoIODAMemory 109
```
jjguerrette, line 16: I'm seeing that my original design of setting job resource requirements uniquely for each `GridDescriptor` is forcing you to copy the same settings in multiple places.  In a follow-on PR, we ought to define mesh-independent job resource requirements in a single place.  This also applies to VerifyObs* resources.

Nothing to change in this PR.

drive.csh
```diff
@@ -130,8 +134,10 @@ cat >! suite.rc << EOF
 {% set firstCyclePoint   = "${firstCyclePoint}" %}
 {% set initialCyclePoint = "${initialCyclePoint}" %}
 {% set finalCyclePoint   = "${finalCyclePoint}" %}
+{% set obtypes = "${obtypes}" %}
```
jjguerrette: Again, something more informative than `obstypes` would be preferred.

PrepJEDI.csh
```diff
@@ -200,33 +205,49 @@ while ( $member <= ${nEnsDAMembers} )
...
-# ABI
-# ===
-ln -sfv $ABIObsDir[$myAppIndex]/${thisValidDate}/abi*_obs_*.h5 ${InDBDir}/
-set ABISUPEROBGRID = $ABISuperOb[$myAppIndex]
-
-# AHI
-# ===
-ln -sfv $AHIObsDir[$myAppIndex]/${thisValidDate}/ahi*_obs_*.h5 ${InDBDir}/
-set AHISUPEROBGRID = $AHISuperOb[$myAppIndex]
+if ( $ArgPrepObsOn == True ) then
```
jjguerrette, line 208: At some point in the future I would prefer that we move this kind of `if` statement to the configuration level to avoid PrepJEDI.csh needing to know where observations are located in these two unique situations.  That being said, this solution works just fine for our first two use-cases.

jjguerrette (changes requested):
Only a few minor changes requested.  Looks great overall!

ObstoIODA.csh
ibanos90: In modeldata.csh we define the variable RDAdataDir which corresponds to the directory were all RDA data is on Cheyenne. This was done in first place because of the script UngribColdStartIC.csh that needs the RDA GFS forecasts. Should we define this variable in other script?

PrepJEDI.csh
ibanos90: Yes, I agree. I think it would be great if all the observations for the different use-cases were under the same folder (variable name), then the specifications could be handled in obsdata.csh for example.

config/mpas/O30kmIE120km/job.csh
ibanos90: I agree. We can put all the same job specifications in a single place.

ObstoIODA.csh
ibanos90: In modeldata.csh. This follows up the previous question.

ObstoIODA.csh
ibanos90: Removed in the last commit

config/ObsPlugs/hofx/sfc.yaml
ibanos90: You are right, I forgot to rename the files. It was updated in the last commit

config/experiment.csh
ibanos90: Sure. I used the suggested name in the last commit

drive.csh
ibanos90: Changed as well.

ObstoIODA.csh
jjguerrette: Thank you for the reminder!  I forgot that you already added other RDA data usage.

config/environmentIODAupgrade.csh
```diff
@@ -0,0 +1,14 @@
...
+module load jedi/gnu-openmpi
+module load json
+module load json-schema-validator
+unalias ecbuild
+#setenv LOCAL_PATH_JEDI_TESTFILES /glade/u/home/maryamao/JEDI_test_files
+git lfs install
+limit stacksize unlimited
+setenv OOPS_TRACE 0  # Note: Some ctests fail when OOPS_TRACE=1
+module list
+setenv GFORTRAN_CONVERT_UNIT 'big_endian:101-200'
```
jjguerrette: Maybe these settings could be defined in a file named `environmentForJedi.csh`, then sourced in `config/environment.csh`.  No need to repeat it twice in two files.

drive.csh
```diff
@@ -165,7 +171,12 @@ cat >! suite.rc << EOF
   {% set PrimaryCPGraph = PrimaryCPGraph + "\\n        CyclingFC:succeed-all => CyclingFCFinished" %}
   {% set PrimaryCPGraph = PrimaryCPGraph + "\\n        CyclingDAFinished" %}
 {% elif CriticalPathType == "Normal" %}
-  {% set PrimaryCPGraph = PrimaryCPGraph + "\\n        CyclingFCFinished[-PT${CyclingWindowHR}H]" %}
+  {# set PrimaryCPGraph = PrimaryCPGraph + "\\n        CyclingFCFinished[-PT${CyclingWindowHR}H]" #}
+  {% if PreprocessObs == "True" %}
+    {% set PrimaryCPGraph = PrimaryCPGraph + "\\n        CyclingFCFinished[-PT${CyclingWindowHR}H] => ObstoIODA" %}
+  {% else %}
+    {% set PrimaryCPGraph = PrimaryCPGraph + "\\n        CyclingFCFinished[-PT${CyclingWindowHR}H]" %}
+  {% endif %}
```
jjguerrette: The commented line 174 can be uncommented and this if statement simplified as follows.
```
{% set PrimaryCPGraph = PrimaryCPGraph + "\\n        CyclingFCFinished[-PT${CyclingWindowHR}H]" %}
{% if PreprocessObs == "True" %}
  {% set PrimaryCPGraph = PrimaryCPGraph + " => ObstoIODA" %}
{% endif %}
```

config/environmentIODAupgrade.csh
ibanos90: You are right. I made the suggested changes.

drive.csh
ibanos90: Thanks for pointing this out and the recommended change!

ObstoIODA.csh
```diff
@@ -154,7 +154,7 @@ end # inst loop
 if ( "${OBTYPES}" =~ *"prepbufr"* ) then
   # Run the ioda-upgrade executable to upgrade to get string station_id and string variable_names
   # ==================
-  source ${mainScriptDir}/config/environmentIODAupgrade.csh
+  source ${mainScriptDir}/config/environment.csh
```
jjguerrette: I recall that you mentioned there was a need to load the the environment without NCO or python in order for the upgrader to work.  Is this supposed to be `environmentForJEDI` instead of `environment`?  If not and only one environment is needed, no need to have two different files.  Maybe we can discuss offline.

If environmentForJEDI needs to exist, then I would think that `${BuildCompiler}` would be derived by sourcing  `config/builds.csh` within `environmentForJEDI.csh` similar to how it is used now in `environment.csh`.  Then the duplication between the two environment scripts would be removed.

ibanos90: Honestly, I am not sure why the upgrader does not work with just `environment.csh` since it contains most of the modules needed. I tried merging both and changing the position of the modules loaded but nothing worked. So, for now, we do need to source `environmentForJEDI.csh` in `ObstoIODA.csh`. I did as you suggested to avoid duplication in the scripts and added a work around to pass `BuildCompiler` as an argument into `environmentForJEDI.csh`. This is because we need to source `environmentForJEDI.csh` inside the Observations folder (`${WorkDir}`). I am not sure if this would the best solution though.

config/environmentForJedi.csh
```diff
@@ -1,8 +1,10 @@
+#!/bin/csh -f
+
```
jjguerrette, line 2: Very nice solution.  Do you mind giving documentation of the argument here (similar to other command-line arguments in the workflow), including possible options?

So far the only possible values are
`gnu-openmpi`
`intel-impi`

ibanos90: Sure. I added a comment for the argument passed and used the same variable name `BuildCompiler`. However, this is also documented in builds.csh. Is it ok to keep both documentation of the argument? Or should I change the variable name?

jjguerrette: It's fine.  The alternative would be to source `config/builds.csh` within `config/environmentForJEDI.csh`.  Up to you.  Let me know when you are satisfied and I will approve.

ibanos90: I agree. The problem with sourcing `config/builds.csh` within `config/environmentForJEDI.csh` is that it needs the whole path to the `config` folder and for that other files need to also be sourced. Then, I thought of adding the argument to simplify. If you think is good for now you can approve it. In the future we can move these changes back once we don't need to run the upgrader for prebufr data.

jjguerrette: Sure.  We'll see what is needed when we have time to revisit.  It's good to get you moving forward.  GNSSRO still in a separate PR?

jjguerrette (approved):

ObstoIODA.csh
```diff
@@ -0,0 +1,187 @@
...
+
+# Process arguments
+# =================
+## args
+# OBTYPES: observation types to pre-process
+set OBTYPES = "$1"
+
+# Setup environment
+# =================
+source config/experiment.csh
```
jjguerrette, line 8: Since config/experiment.csh is sourced, I think the OBSTYPES argument is redundant.

jjguerrette (changes requested):

drive.csh
```diff
@@ -130,8 +134,10 @@ cat >! suite.rc << EOF
 {% set firstCyclePoint   = "${firstCyclePoint}" %}
 {% set initialCyclePoint = "${initialCyclePoint}" %}
 {% set finalCyclePoint   = "${finalCyclePoint}" %}
+{% set preprocessObsList = "${preprocessObsList}" %}
```
jjguerrette: Thus, this line can be removed as well.

ObstoIODA.csh
ibanos90: You are right. I will remove the argument. Thanks for catching that!

drive.csh
ibanos90: Sure. Thank you!

ObstoIODA.csh
jjguerrette: Then lines 6-10 can be removed entirely, right?

jjguerrette (approved):

ObstoIODA.csh
ibanos90: Yes, sorry I thought I had deleted them.

PR #85, opened 2022-03-23
Update initial sea surface fields

### LINKED ISSUE #84: Update sea surface for cold start initialization
Update the initial condition with the SST and XICE from the GFS analysis valid at the initial condition time.

ibanos90:
### Description
This PR updates the initial condition with the SST and XICE from the GFS analysis valid at the initial condition time. This is added for the ColdStart initialization type, therefore it can be used for any date and the real time run. The main change is in `forecast.csh` where the initial condition file is updated instead of the forecast output files. In addition, the observation processing dependencies were adjusted so that they can be downloaded and converted to IODA outside the forecast- and DA-dependent cycling.

### Issue closed
Closes https://github.com/NCAR/MPAS-Workflow/issues/84

### Tests completed
 - [x] eda_OIE120km_WarmStart

### DIFF
--- a/config/ColdStartModelData.csh
+++ b/config/ColdStartModelData.csh
@@ -17,8 +17,8 @@ set GFSAnaDirInner = ${InnerModelData}/GFSAna
 set GFSAnaDirEnsemble = ${EnsembleModelData}/GFSAna
 
 setenv InitICWorkDir ${GFSAnaDirOuter}
-# TODO: enable sea-surface file generation, then turn on sea-surface updating
-setenv updateSea 0
+setenv SeaFilePrefix ${InitFilePrefixOuter}
+
 if ( "$DAType" !~ *"eda"* ) then
   setenv StaticFieldsDirOuter ${GFSAnaDirOuter}/${FirstCycleDate}
   setenv StaticFieldsDirInner ${GFSAnaDirInner}/${FirstCycleDate}
--- a/drive.csh
+++ b/drive.csh
@@ -154,7 +154,10 @@ cat >! suite.rc << EOF
     [[[${cyclingCycles}]]]
       graph = '''{{PrimaryCPGraph}}{{SecondaryCPGraph}}
   {% if InitializationType == "ColdStart" %}
-        GenerateColdStartIC => CyclingFC
+        # ensure that there is a GFS analysis file valid at the analysis time
+        # from which to pull sea-surface fields
+        {{GFSAnalysisWorkflow}} => CyclingFC
+        # get observations and convert them to IODA outside the cycling
         GetObs => ObsToIODA
         ObsToIODA => InitCyclingDA
   {% endif %}
--- a/drive.csh
+++ b/drive.csh
@@ -124,7 +124,6 @@ cat >! suite.rc << EOF
 {% set EnsVerifyMembers = range(1, nEnsDAMembers+1, 1) %}
 # Cold initial conditions from GFS analysis
 {% set GFSAnalysisWorkflow = "GetGFSanalysis => UngribColdStartIC => GenerateColdStartIC" %}
-
 [cylc]
   UTC mode = False
   [[environment]]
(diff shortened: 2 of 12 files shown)


config/ColdStartModelData.csh
```diff
@@ -17,8 +17,8 @@ set GFSAnaDirInner = ${InnerModelData}/GFSAna
 set GFSAnaDirEnsemble = ${EnsembleModelData}/GFSAna
 
 setenv InitICWorkDir ${GFSAnaDirOuter}
-# TODO: enable sea-surface file generation, then turn on sea-surface updating
-setenv updateSea 0
```
jjguerrette, line 21: I still want `updateSea` to be a configurable parameter.  That can live in `config/modeldata.csh` as before until we extend the yamls to these settings.

drive.csh
```diff
@@ -159,6 +153,11 @@ cat >! suite.rc << EOF
 ## Critical path for cycling
     [[[${cyclingCycles}]]]
       graph = '''{{PrimaryCPGraph}}{{SecondaryCPGraph}}
+  {% if InitializationType == "ColdStart" %}
+        GenerateColdStartIC => CyclingFC
```
jjguerrette: Should this be corrected as follows?

`GenerateColdStartIC => CyclingFC` --> `{{GFSAnalysisWorkflow}}  => CyclingFC`

I guess L157 is needed in order to ensure there is an analysis file from which to pull sea-surface fields.  Do you mind adding a comment stating as much?  Then later we may decide to generate "sfc_update" files instead of using a full analysis and we can make the proper substitution here (no comment needed for that info).

jjguerrette (changes requested):
Looks good.  Only minor requests.

Thank you for adjusting the observation retrieval dependencies.  Would be good to include that in the PR message.

config/ColdStartModelData.csh
ibanos90: Ok. I re-added the definition of `updateSea` back to `config/modeldata.csh` and the if statement in `forecast.csh`

drive.csh
ibanos90: The line was corrected and some comments added. Thanks for pointing this out!

drive.csh
```diff
@@ -154,7 +154,10 @@ cat >! suite.rc << EOF
     [[[${cyclingCycles}]]]
       graph = '''{{PrimaryCPGraph}}{{SecondaryCPGraph}}
   {% if InitializationType == "ColdStart" %}
-        GenerateColdStartIC => CyclingFC
+        # ensure that there is a GFS analysis file valid at the analysis time
+        # from which to pull sea-surface fields
+        {{GFSAnalysisWorkflow}} => CyclingFC
+        # get observations and convert them to IODA outside the cycling
```
jjguerrette: Slight adjustments to comments and indentation below
```
      # critical path tasks that are independent of the previous forecast
  {% if InitializationType == "ColdStart" %}
      # ensure that there is a GFS analysis file valid at the analysis time
      # from which to pull sea-surface fields
      {{GFSAnalysisWorkflow}} => CyclingFC
      # get observations and convert them to IODA
      GetObs => ObsToIODA
      ObsToIODA => InitCyclingDA
  {% endif %}
```
2 points for follow-on PRs.  Do you mind putting the list of follow-on tasks in a GitHub issue?  We can work on them together.

The `ObsToIODA => InitCyclingDA` part is not valid when CriticalPathType == "Bypass".  It think it's time to group some of the Variational-type cylc tasks together in a similar way to what you have done for `GFSAnalysisWorkflow`, but still need to think it through.

Using `InitializationType == "ColdStart"` as the logical check here is only correct for the sea-surface update, not the obs processing, which depends on `$observations__resource` only.

ibanos90: Sure. I added these points to the list and will create the Issue with all of them.

jjguerrette (approved):

PR #91, opened 2022-03-31
Utilize surface stream for SST and XICE updating

### LINKED ISSUE #90: Avoid file duplication from ncks step for SeaSurface update
https://github.com/NCAR/MPAS-Workflow/blob/90b2da03ab6a8738ab1cb4fde72e20418c45b307/forecast.csh#L87

The line above and a later ncks command create a copy of a file that was previously linked into the forecast directory.  Likely ncks cannot append to the originally linked file.  This increases file storage space for model states by 50%, i.e., (1)forecast/background, (2) analysis/IC, (3) IC w/ sst.  (3) is an extra file, when (2) could have the sst and xice updated and then no extra file is needed.  This added expense will be most noticeable for the 60km EDA.

A quick solution is to modify file (2) instead of generating file (3).

The best solution is to read sst and xice directly during the forecast via the surface stream, and then skip the ncks step altogether.

jjguerrette:
### Description
The sea-surface updating is carried out via the surface stream during the forecast execution instead of using the `ncks` command before the forecast.  The `updateSea` and forecast `precision` are moved to the new `forecast.yaml` base scenario configuration.  Other forecast-specific settings can be moved there eventually.

The `GraphInfoDir` setting is moved from `modeldata.csh` to the `model.yaml`  base scenario configuration.

### Issue closed

Closes #90

### Tests completed
 - [x] eda_OIE120km_WarmStart

### DIFF
--- a/GenerateColdStartIC.csh
+++ b/GenerateColdStartIC.csh
@@ -4,6 +4,7 @@ date
 
 # Setup environment
 # =================
+source config/forecast.csh
 source config/model.csh
 source config/filestructure.csh
 source config/modeldata.csh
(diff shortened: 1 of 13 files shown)


ibanos90 (approved):
Thanks @jjguerrette for adding this capability. I ran the tests and all completed successfully.

PR #92, opened 2022-04-01
Quality control flags in YAML stubs and VarBC for AMSUA

ibanos90:
### Description
This PR is a follow up of #89 in which the feature branch is renamed and changes are made accordingly following that PR's comments/suggestions.

Here, the variable `biasCorrection` gets only the values `True` or `False` indicating that bias correction is not applied for pre-generated periods (PANDACArchive) since the observations are already bias corrected, and that variational bias correction is applied otherwise. The `biasCorrection` settings is available in both `variational` and `hofx` appications, but is only implemented so far for `variational`.  Setting `biasCorrection` to True for `variational` also turns on variational bias correction coefficient updating.  That new functionality enables the writing of the satellite bias correction coefficients at each cycle and uses the coefficients calculated from the previous cycle as prior estimates.

In addition, the YAML stubs are split and each part is added to a corresponding folder (base, filters, bias, filtersWithBias) for variational and hofx applications. In this way, a great flexibility is allowed in the sense of constructing the full application YAML from a modular set of sub-yamls.

Additionally, `satwnd` observations are enabled for both `variational` and `hofx` applications.  `satwnd` observations will only be used when `observations.resource` is set to either of `GladeRDAOnline` or `NCEPFTPOnline`, such that observations are converted to IODA format from BUFR.

### Issue closed
Closes #43
Closes #82
Closes #86

### Tests completed
 - [x] 3dvar_OIE120km_WarmStart
 - [x] 3denvar_OIE120km_WarmStart
 - [x] 3dvar_OIE120km_ColdStart
 - [x] 3denvar_O30kmIE60km_WarmStart
 - [x] eda_OIE120km_WarmStart

PR #96, opened 2022-04-08
Fix channel selection for NCDIAG AMSUA variational

### LINKED ISSUE #95: Wrong channel selection for NCDIAG AMSUA variational
For example:
https://github.com/NCAR/MPAS-Workflow/blob/b13210ac8103f738133ff67ded1df8b129f08acf/config/ObsPlugs/variational/base/amsua_n19.yaml#L9

Previously we only assimilated a limited channel set for AMSUA instruments in PANDACArchive experiments.  We could remove some of the channels with `filters` similar to what is done in the `filtersWithBias` yamls.

ibanos90:
### Description
This PR fixes a bug in the channel selection for NCDIAG AMSUA variational. In order to account for the limited channel set for AMSUA instruments used in `PANDACArchive` experiments, the `Useflag check` was added to the base filters. This allows to use the same base yaml files for experiments using `PANDACArchive` or `*Online` data. The channel selection in the check follows the used before the last PR #92.

### Issue closed
Closes #95

### Tests completed
 - [x] eda_OIE120km_WarmStart

### DIFF
--- a/config/ObsPlugs/variational/filters/amsua_aqua.yaml
+++ b/config/ObsPlugs/variational/filters/amsua_aqua.yaml
@@ -6,3 +6,19 @@
   - filter: Background Check
     threshold: 3.0
     <<: *multiIterationFilter
+#  Useflag check
+  - filter: Bounds Check
+    filter variables:
+    - name: brightness_temperature
+      channels: *amsua_aqua_channels
+    test variables:
+    - name: ChannelUseflagCheckRad@ObsFunction
+      channels: *amsua_aqua_channels
+      options:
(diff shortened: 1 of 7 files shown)


jjguerrette (approved):
Thank you for the quick PR.  This is exactly what I requested in the issue.  In a later PR I would like us to figure out a way to make the channel selection much easier to read and modify.  The `ChannelUseflagCheckRad@ObsFunction` is not user-friendly.

ibanos90:
> 

That would be great. I will add it to the follow up tasks. Thanks.

PR #98, opened 2022-04-12
Activate Github actions

mer-a-o:
### Description
This PR adds a yaml file for github actions with simple instructions to start a cycle on a self-hosted runner (in this case Cheyenne) with `run-test` label. The main goal here is to add this yaml and expand it in the future.

### DIFF
--- a/.github/workflows/cycle.yml
+++ b/.github/workflows/cycle.yml
@@ -1,0 +1,36 @@
+# This is a basic workflow to help you get started with Actions
+
+name: CI
+
+# Controls when the workflow will run
+on:
+  # Triggers the workflow on push or pull request events but only for the develop branch
+  push:
+    branches: [ develop ]
+  pull_request:
+    branches: [ develop ]
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  # This workflow contains a single job called "build"
+  build:
+    # The type of runner that the job will run on
+    runs-on: ubuntu-latest
+
+    # Steps represent a sequence of tasks that will be executed as part of the job
+    steps:
+      # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
+      - uses: actions/checkout@v3
+
+      # Runs a single command using the runners shell
+      - name: Run a one-line script
+        run: echo Hello, world!
+
+      # Runs a set of commands using the runners shell
+      - name: Run a multi-line script
+        run: |
+          echo Add other actions to build,
+          echo test, and deploy your project.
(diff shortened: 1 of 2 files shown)


.github/workflows/cycle.yml
```diff
@@ -0,0 +1,41 @@
...
+      - name: checkout
+        uses: actions/checkout@v3
+
+      - name: run-mpas-workflow
+        run: |
+          source /etc/profile.d/modules.sh
+          source ~/.bashrc
+          export JEDI_OPT=/glade/work/jedipara/cheyenne/opt/modules
+          module use $JEDI_OPT/modulefiles/core
+          module load jedi/gnu-openmpi
```
jjguerrette: Are these environment settings needed for some aspect on your end of the testing?

They are not typically needed to initiate an MPAS-Workflow scenario suite.  The compiler/run environment is loaded individually for each cylc task in its job script (e.g., config/environmentForJedi.csh).

.github/workflows/cycle.yml
```diff
@@ -0,0 +1,41 @@
...
+    types: [labeled]
+    branches: [develop]
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  Run-mpas-workflow:
+    if: ${{ github.event.label.name == 'run-cycle' }}
```
jjguerrette: Do you mind changing `run-cycle` to `run-test` to match the name of `test.csh`? EDIT: or to match `test.yaml`, I suppose.

.github/workflows/cycle.yml
mer-a-o: If they are not needed we can remove them. We only need to keep `source /etc/profile.d/modules.sh` if we have `module load` commands in the code.

.github/workflows/cycle.yml
mer-a-o: sure

.github/workflows/cycle.yml
```diff
@@ -0,0 +1,34 @@
+# This is a basic workflow to help you get started with Actions
+
+name: test-mpas-workflow
+
+# Controls when the workflow will run
+on:
+  # uncomment if we decide to run tests when merging to develeop branch
```
jjguerrette: `develeop` -> `develop`

jjguerrette (approved):

jjguerrette:
Oh maybe rename `cycle.yml` to `test.yml` also.

jjguerrette (approved):

jjguerrette:
Depending on what you envision going in the `workflows` directory in the future, I might propose that it be renamed to `gh-actions` or something similar in a follow-on PR.  Looks great to merge this one.

jjguerrette:
Disregard my last comment.  I just noticed `workflows` is already under `.github`.  That will work great.

jjguerrette:
@mer-a-o, is this ready to merge?

mer-a-o:
@jjguerrette yes, it's ready.

PR #105, opened 2022-05-05
Updates for the IAU

syha:
### Description
New implementation for IAU in MPAS-JEDI

# How to use
In scenarios/3denvar_OIE120km_WarmStart_IAU.yaml
```
forecast:
  IAU: True
```
These two lines demonstrate how to enable IAU. The default value (False) is located in scenarios/base/forecast.csh.

### List of files changed and added
M       SetupWorkflow.csh
M       config/applications/forecast.csh
M       config/mpas/forecast/streams.atmosphere
M       config/tools.csh
M       config/workflow.csh
M       drive.csh
M       forecast.csh
M       include/tasks.rc
M       include/verification.rc
M       run.csh
A       runs/120km3denvarIAU.yaml
A       scenarios/3denvar_OIE120km_WarmStart_IAU.yaml
M       scenarios/base/forecast.yaml
M       scenarios/base/workflow.yaml
A       tools/TimeFmtChange.py
A       tools/create_amb_in_nc.py

### Tests completed
 - [x] eda_OIE120km_WarmStart


liujake:
@syha Conflicts needs to be resolved first. I remember you mentioned you fixed some IAU in MPAS model. If so, you also need to make a PR to https://github.com/JCSDA-internal/MPAS-Model repo's 'release-stable' branch.

SetupWorkflow.csh
```diff
@@ -86,6 +88,20 @@ sed -e 's@WorkDirsTEMPLATE@CyclingFCDirs@' \
     forecast.csh > ${JobScript}
 chmod 744 ${JobScript}
 
+else ## CyclingFC for IAU
+
+echo "Making CyclingFC job script for IAU"
+set JobScript=${mainScriptDir}/CyclingFC.csh
+sed -e 's@WorkDirsTEMPLATE@CyclingFCDirs@' \
+    -e 's@StateDirsTEMPLATE@CyclingDAOutDirs@' \
+    -e 's@CyclingFrequencyHR@'${CyclingWindowHR}'@' \
```
jjguerrette: `MPAS-Workflow/forecast.csh` already has access to `CyclingWindowHR` on its L25:
```
source config/workflow.csh
```
So the `CyclingFrequencyHR` variable is redundant and can be removed from this sed substitution.

config/experiment.csh
```diff
@@ -228,14 +227,27 @@ setenv nOuterIterations ${#nInnerIterations}
...
+# With config_IAU_option = 'on' in namelist.atmosphere,
+# forecast is no longer initialized at the analysis time,
+# but is shifted to -3h and integrated for 9h [IAUfcLengthHR] until the next analysis time.
+# Hence, the IAU setting only affects forecasts in the model, not the jedi analysis.
+# For this configuration, 1) we need to print out the model output every 3h and
+# 2) compute the analysis increments at the analysis time in a netcdf file
+#    (using tools/create_amb_in_nc.py).
+setenv IAU True 			# OPTIONS: True/False
+setenv IAUoutIntervalHR 3               # forecast output interval between CyclingDA analyses
+setenv IAUfcLengthHR 9                  # forecast length with IAU on: [-3h, 6h]
```
jjguerrette: As we discussed offline, these three parameters will get moved to `scenarios/base/forecast.csh`.  If `IAUoutIntervalHR` is always equal to `IAUfcLengthHR - CyclingWindowHR`, then maybe only two new variables are needed for `IAU`.  It's good to give users a fool-proof control knob instead of two redundant knobs.  Then in `config/applications/forecast.csh`, you can have something like this:
```
$setLocal IAU
$setLocal IAUfcLengthHR

@IAUoutIntervalHR = $IAUfcLengthHR - $CyclingWindowHR
setenv IAUoutIntervalHR $IAUoutIntervalHR
```

Alternatively, you might opt to make `IAUfcLengthHR` hard-coded in the script and invisible to users.  Up to you.

The main challenge in getting this to work with the current develop branch is that `SetupWorkflow.csh` no longer sources `config/applications/forecast.csh` or any other application-specific config scripts.  There is a convoluted way to get `forecast`  settings up to `SetupWorkflow.csh` via `generateExperimentConfig.csh`, but it is a pathway I would like to avoid.  I'm working toward avoiding most of the sed substitution in `SetupWorkflow.csh`, because it is non-transparent and error-prone.

Instead I would recommend creating new settings in `config/applications/forecast.csh` like the following:
```
if ($IAU == True) then
  setenv FCLengthHR $IAUfcLengthHR
  setenv FCOutIntervalHR $IAUoutIntervalHR
else
  setenv FCLengthHR $CyclingWindowHR
  setenv FCOutIntervalHR $CyclingWindowHR
endif
```

Then I would like us to change the forecast output interval and forecast length from being controlled via sed substitution in `SetupWorkflow.csh` to command-line arguments fed directly to the `forecast.csh` task script (i.e., `MPAS-Workflow/forecast.csh`).

That involves modifying these files:
`drive.csh`
`SetupWorkflow.csh`
`include/tasks.rc`
`forecast.csh`
`config/workflow.csh`

E.g., for `include/tasks.rc`:
```
diff --git a/include/tasks.rc b/include/tasks.rc
index edd26c2..426f3d1 100644
--- a/include/tasks.rc
+++ b/include/tasks.rc
@@ -203,7 +203,7 @@
 {% for mem in allMembers %}
   [[ForecastMember{{mem}}]]
     inherit = Forecast
-    script = $origin/Forecast.csh "{{mem}}"
+    script = $origin/Forecast.csh "{{mem}}" "{{FCOutIntervalHR}}" "{{FCLengthHR}}"
     [[[job]]]
       execution retry delays = {{CyclingFCRetry}}
 {% endfor %}
@@ -262,7 +262,7 @@
       -l = select=1:ncpus=36:mpiprocs=36
   [[ExtendedMeanFC]]
     inherit = ExtendedFCBase
-    script = $origin/ExtendedMeanFC.csh "1"
+    script = $origin/ExtendedMeanFC.csh "1" "{{ExtendedFCOutIntervalHR}}" "{{ExtendedFCLengthHR}}"
   [[HofXMeanFC]]
     inherit = HofXBase
   [[VerifyModelMeanFC]]
@@ -371,7 +371,7 @@
   # ensemble of extended forecasts from ensemble of analyses
   [[ExtendedFC{{mem}}]]
     inherit = ExtendedEnsFC
-    script = $origin/ExtendedEnsFC.csh "{{mem}}"
+    script = $origin/ExtendedEnsFC.csh "{{mem}}" "{{ExtendedFCOutIntervalHR}}" "{{ExtendedFCLengthHR}}"
   [[HofXEnsFC{{mem}}]]
     inherit = HofXBase
   [[VerifyModelEnsFC{{mem}}]]
```

In `drive.csh`, you would change the following:
```
diff --git a/drive.csh b/drive.csh
index 7d4eb3d..d59ecfe 100755
--- a/drive.csh
+++ b/drive.csh
@@ -93,8 +93,10 @@ cat >! suite.rc << EOF
 {% set FC2DAOffsetHR = "${FC2DAOffsetHR}" %}
 {% set ExtendedMeanFCTimes = "${ExtendedMeanFCTimes}" %}
 {% set ExtendedEnsFCTimes = "${ExtendedEnsFCTimes}" %}
-{% set ExtendedFCWindowHR = ${ExtendedFCWindowHR} %} #integer
-{% set ExtendedFC_DT_HR = ${ExtendedFC_DT_HR} %} #integer
+{% set FCOutIntervalHR = ${FCOutIntervalHR} %} #integer
+{% set FCLengthHR = ${FCLengthHR} %} #integer
+{% set ExtendedFCOutIntervalHR = ${ExtendedFCOutIntervalHR} %} #integer
+{% set ExtendedFCLengthHR = ${ExtendedFCLengthHR} %} #integer
```
With these name changes implied:
`ExtendedFCWindowHR`-->`ExtendedFCLengthHR`
`ExtendedFC_DT_HR`-->`ExtendedFCOutIntervalHR`


What do you think?

config/job.csh
```diff
@@ -8,7 +8,7 @@ source config/experiment.csh
 
 ## *AccountNumber
 # OPTIONS: NMMM0015, NMMM0043
-setenv StandardAccountNumber NMMM0043
+setenv StandardAccountNumber NMMM0013
```
jjguerrette: Note to self: revert before merge.  Fine for now.

getCycleVars.csh
```diff
@@ -8,6 +8,9 @@ set prevCycleDate = `$advanceCYMDH ${thisCycleDate} -${CyclingWindowHR}`
 setenv prevCycleDate ${prevCycleDate}
 #setenv nextCycleDate ${nextCycleDate}
 
+set IAUDate = `$advanceCYMDH ${thisCycleDate} -${IAUoutIntervalHR}`
```
jjguerrette: Since `IAUDate` is only used by `forecast.csh`, this definition can be moved to that script.

jjguerrette (commented):
The changes to `forecast.csh` should be fine for a first take.  We discussed offline whether the cost of the AmB step will be a problem.  I think in a later iteration of IAU, we will write AmB directly from the Variational application.  So it's not worth spending time on optimizing that aspect right now.

We just need to work on how the new IAU settings are propagated to the workflow.  The workflow is configured via yaml, which is very different from this several month old version (so old... :-)).

Thank you for your patience on the evolving workflow.

jjguerrette:
Another detail I forgot is that since we use `forecast.csh` for both the critical path Forecast tasks and the ExtendedFC tasks, we need to have another argument that tells the script either the IC offset time or whether IAU is being used.  It might be possible to avoid that extra argument by using the `DA2FCOffsetHR` and `FC2DAOffsetHR` settings that currently reside in `config/workflow.csh`, but that requires additional changes.  I did not expect `DA2FCOffsetHR` to be negative, and in your case I think it is `-3`.

SetupWorkflow.csh
syha: Thanks for your patience. I'm back to this now, and agree with you on the simplification.

config/experiment.csh
syha: In terms of defining IAUoutIntervalHR, I agree with you on deriving it from CyclingWindowHR.
In fact, even IAUfcLengthHR can be derived from it. 
Let's say CyclingWindowHR as dt.
Then, IAUoutIntervalHR = dt/2 and IAUfcLengthHR = 3*IAUoutIntervalHR, no matter what dt is.
Where would be the place to define these?

With this set-up, all users need to specify is CyclingWindowHR and IAU (e.g., T or F) in the yaml.
And we also need to be able to edit streams.atmosphere for the output interval accordingly.

getCycleVars.csh
syha: Fine with me.

config/experiment.csh
jjguerrette: Those math operations and "derived" settings would go in config/applications/forecast.csh.

scenarios/3denvar_OIE120km_WarmStart_IAU.yaml
```diff
@@ -0,0 +1,15 @@
...
+observations:
+  resource: PANDACArchive
+model:
+  outerMesh: 120km
+  innerMesh: 120km
+  ensembleMesh: 120km
+variational:
+  DAType: 3denvar
+forecast:
+  IAU: True
```
jjguerrette, line 24: I see why the job submission failed for you.  You may check this file to see yourself:

/glade/scratch/syha/cylc-run/syha_3denvar-60-iter_OIE120km_WarmStart_IAUTesting/log/job 20180415T0000-06/DAMember1/01/job-activity.log

Please add the following in order to use accounts to which you have access:
```yaml
job:
  CPAccountNumber: NMMM0015
  NCPAccountNumber: NMMM0015
  SingleProcAccountNumber: NMMM0015
```

You may substitute any of the account numbers you have access to, although the `SingleProcAccountNumber` is used to submit jobs on Casper.  Not all accounts have access to casper nodes.

config/builds.csh
```diff
@@ -56,15 +56,15 @@ setenv InitBuildDir ${commonBuild}/bin
 setenv ForecastTopBuildDir ${commonBuild}
 
 # Use a static single-precision build of MPAS-A to conserve resources
-setenv ForecastBuildDir /glade/p/mmm/parc/liuz/pandac_common/20220309_mpas_bundle/code/MPAS-gnumpt-single
-setenv ForecastEXE ${MPASCore}_model
+#setenv ForecastBuildDir /glade/p/mmm/parc/liuz/pandac_common/20220309_mpas_bundle/code/MPAS-gnumpt-single
+#setenv ForecastEXE ${MPASCore}_model
```
jjguerrette: Note to self: revert before merging

forecast.csh
```diff
@@ -28,7 +31,7 @@ source config/tools.csh
 source config/model.csh
 source config/modeldata.csh
 source config/builds.csh
-source config/environmentMPT.csh
+source config/environmentJEDI.csh
```
jjguerrette: Note to self: revert before merging

forecast.csh
```diff
@@ -181,30 +229,25 @@ if ( ${self_fcLengthHR} == 0 ) then
...
   # Run the executable
   # ==================
-  rm ./${ForecastEXE}
+  if( -e ${ForecastEXE} ) rm ./${ForecastEXE}
   ln -sfv ${ForecastBuildDir}/${ForecastEXE} ./
   # mpiexec is for Open MPI, mpiexec_mpt is for MPT
-  #mpiexec ./${ForecastEXE}
-  mpiexec_mpt ./${ForecastEXE}
+  mpiexec ./${ForecastEXE}
+  #mpiexec_mpt ./${ForecastEXE}
```
jjguerrette: Note to self: revert before merging

scenarios/3denvar_OIE120km_WarmStart_IAU.yaml
```diff
@@ -0,0 +1,15 @@
...
+builds:
+  commonBuild: /glade/scratch/syha/mpas-bundle-iaufix-build
+  #commonBuild: /glade/work/syha/mpas-bundle-build-gnu-openmpi #default
```
jjguerrette: Note to self: resolve before merging.  Will need a new build for all scenarios in order to use IAU.

getCycleVars.csh
```diff
@@ -10,6 +10,9 @@ set prevCycleDate = `$advanceCYMDH ${thisCycleDate} -${CyclingWindowHR}`
...
+setenv IAUDate ${IAUDate}
+
```
jjguerrette: Revert these changes.  `IAUDate` is set in `/forecast.csh`, which is the only application that uses it.

tools/TimeFmtChange.py
```diff
@@ -0,0 +1,27 @@
...
+import datetime as dt
+import argparse
+
+def TimeFmtChange():
+
+  # Parse command line
+  ap = argparse.ArgumentParser()
+  ap.add_argument('date', type=str,
+                  help='Date in YYYYMMDDHH format')
+  ap.add_argument('fmt', default=0, type=int, nargs = '?')
```
jjguerrette, line 14: I like this new tool.  Can you make the `fmt` argument take a string instead of an int?  Then the format (e.g., `"%Y-%m-%d_%H:00:00"`) can simply be provided on the command line.

Maybe the input and output formats need to be provided.  That would be more transparent than the 0 and 1 options.  We could also use this tool for all of the other date format conversions instead of using shell commands.  Note that I am not asking you to update all of our other date format conversions.

jjguerrette (commented):
Just because we scraped together the changes in rather involved conversations, I wanted to remind us of the places that need to be reverted before I forget.  No need to address any of these comments until after the IAU is working for the whole month.

config/builds.csh
jjguerrette: Actually, not revert, but need to update the single-precision MPT build with @syha's latest changes.

jjguerrette:
@syha, when you feel that this PR is functionally ready, let's take the final steps needed to merge it in.  We can merge even if the verification results do not show outright improvement with IAU, just to get the changes in and allow for some other refactoring I have coming without affecting your work.

jjguerrette:
@syha: after the tests complete, the last remaining action item for me is to create a new build that includes the latest changes to MPAS-Model.  I will push the directory name up to this branch (`scenarios/base/builds.yaml`).  Otherwise this PR is approved and ready to merge.

liujake:
Better document a bit in the PR message: IAU is on or off by default? How to turn it on/off in the setting.

scenarios/3denvar_OIE120km_WarmStart_IAU.yaml
```diff
@@ -0,0 +1,30 @@
...
+model:
+  outerMesh: 120km
+  innerMesh: 120km
+  ensembleMesh: 120km
+experiment:
+  ExpSuffix: '_IAUfix2'
+variational:
+  DAType: 3denvar
+forecast:
+  IAU: True
```
jjguerrette, line 24: These two lines demonstrate how to enable IAU.  The default value (`False`) is located in `scenarios/base/forecast.csh`.

liujake (approved):
Suggest merge if it is ready. IAU is off by default so should not affect existing experimental settings.

jjguerrette (approved):

jjguerrette:
@syha, note that I will delete your branch on github because it is the same as develop now.

PR #106, opened 2022-05-06
Update to 04MAY2022 build

### LINKED ISSUE #104: Update to 04MAY2022 build
More updates to speed up GetValues, and to reduce the number of observation feedback files along with associated disk space.  Changes to observation feedback file number may require modification to diagnostics scripts.

In progress: feature/04MAY2022

jjguerrette:
### Description
Use a recent build that includes the new IODA IO pool feature, which allows for the ObsSpace writer to use a subset of the total processors for writing.  Currently each writer process generates a single file.  Previously each process associated with a simulation produced its own file, e.g., 36 files per observation type corresponding to 36 processors in our 120km HofX.  ObsSpace file storage is reduced as follows for files with the `obsout_` prefix produced by HofX in the first cycle of the 2018 experiment:
- Before IO pool feature (36pe): 459MB
- IO pool, default max pool size of 10: 240MB
- IO pool, max pool size of 2: 182MB

Timings for oops::ObsSpace::save, i.e., the writer:
- Before IO pool feature (36pe): 5.4s
- IO pool, default max pool size of 10: 7.9s
- IO pool, max pool size of 2: 25.4s

The GeoVaLs and ObsDiagnostics files still take up a large amount of disk space, but are only produced for a subset of radiance instruments in order to perform cloud-dependent binning.  If we only saved the the ObsSpace files, keeping them around for 80 members over 120 cycles would cost 1.7TB with a max pool size of 2, a 4x reduction over the previous IODA code.  For EDA experiments, it would be reasonable to turn off the cloud-dependent binning that uses GeoVaLs and ObsDiagnostics, then save the ObsSpace files.

### Issue closed

Closes #104 

### Tests completed
 - [x] eda_OIE120km_WarmStart


liujake (approved):

ibanos90 (approved):
All tests completed successfully.

PR #109, opened 2022-05-25
Adding IASI observations to MPAS-Workflow

### LINKED ISSUE #57: Add IASI observations to MPAS-Workflow
Enhance the benchmark observations in current MPAS-Workflow to assimilate clear-sky IASI observations. Will start with Emily Liu’s IASI QC from GDAS yamls:
https://github.com/JCSDA-internal/ufo/blob/develop/ewok/jedi-gdas/iasi_metop-a.yaml
https://github.com/JCSDA-internal/ufo/blob/develop/ewok/jedi-gdas/iasi_metop-b.yaml

ibanos90:
### Description
This PR enables the assimilation of clear-sky IASI observations into the MPAS-Workflow. The corresponding yaml stubs were added following Emily Liu’s IASI QC from GDAS yamls: [https://github.com/JCSDA-internal/ufo/blob/develop/ewok/jedi-gdas/iasi_metop-a.yaml](url). Two methods were implemented in `mpas-jedi` to determine the tropopause pressure needed in the the cloud detection filter for infrared observations. These were the Thompson and WMO methods adapted from the` FV3-JEDI` and `WRFDA` codes, respectively (PR [#717](https://github.com/JCSDA-internal/mpas-jedi/pull/717)). The default method used in the workflow is Thompson. It was also added the variational bias correction with the capability to use the most recent previous satellite bias correction coefficients created due to the existence of days with no IASI observations in RDA archives during 2018. Also, the ability to not assimilate the IASI observation if an observation file contains missing channel indices has been implemented. A new scenario was added to the workflow for the assimilation of IASI observations. A yaml containing the default built code was added to the workflow, which allows to easily change built codes between the experiments. From the 616 channels available in the observations files, only a subset of 174 channels is actually assimilated following GSI [global_satinfo](https://www.nco.ncep.noaa.gov/pmb/codes/nwprod/gfs.v16.1.7/fix/fix_gsi/global_satinfo.txt). However, since all observations are currently kept during the data assimilation (whether monitored or not), the list with only 174 channels was made the default for now. 

IASI observations were downloaded from RDA archives on Cheyenne and converted to the IODA-v2 format using Jamie's obs2ioda converter for the 2018 and 2020 predefined period outside of the workflow. There are available hourly and six hourly files for 2018 and hourly files for 2020. The initial coefficients for the variational bias correction were also created offline for both periods.

### Issue closed
Closes #57

### Tests completed
 - [x] eda_OIE120km_WarmStart
 - [x] IASI120km3denvar


PrepJEDI.csh
```diff
@@ -213,8 +214,19 @@ foreach instrument ($observations)
...
+  ln -sfv ${obsFile} ${InDBDir}/${instrument}_obs_${thisValidDate}.h5
+
+  # for radiance observations (iasi for now)
+  # check if any channel index is missing (== -999)
+  if ( -e ${obsFile} && "${instrument}" =~ *"iasi"* ) then
+    set missingChannels = `$checkMissingChannels ${obsFile}`
+    if ( ${missingChannels} == True ) then
+      # remove the data
+      echo "WARNING: removing ${instrument} due to missing value in channel indices"
+      rm -rf ${InDBDir}/${instrument}_obs_${thisValidDate}.h5
```
jjguerrette: Only `rm` is needed, not `rm -rf`.

PrepJEDI.csh
```diff
@@ -283,6 +295,19 @@ foreach instrument ($observations)
   foreach i ($instrumentsAllowingBiasCorrection)
     if ("$instrument" == "$i") then
       set allowsBiasCorrection = True
+      if ( ${thisValidDate} != ${nextFirstDate} && ! -e ${CyclingDAWorkDir}/$prevValidDate/dbOut/satbias_${i}.h5 ) then
+          set dateListback = (`$dateList ${FirstCycleDate} ${thisCycleDate} ${self_WindowHR}`)
+          foreach dt (${dateListback})
+            if ( ! -e ${CyclingDAWorkDir}/${dt}/dbOut/satbias_${i}.h5 && ${dt} != ${thisValidDate}) then
```
jjguerrette: The first if clause is a little confusing to me, and likely would be to anybody revisiting the script later.  It appears to be checking whether the satellite bias file for `dt` (date-time?) does not exist AND whether `dt` does not equal `thisValidDate`.  If both conditions are met, then continue.  Why would those two conditions together warrant a continuation?

Possibly you would like some functionality like below? Though I'm not sure what the overall goal is.


```
set dateListback = (`$dateList ${FirstCycleDate} ${thisCycleDate} ${self_WindowHR}`)
set found=False
foreach dt (${dateListback})
  # check for satbias file at dt
  if ( -e ${CyclingDAWorkDir}/${dt}/dbOut/satbias_${i}.h5 ) then
    set biasCorrectionDir = ${CyclingDAWorkDir}/${dt}/dbOut
    set found=True
    break
  endif
end
# if no online updated satbias files exist, use first cycle's satbias file
if ($found == False) then
  set biasCorrectionDir = $initialVARBCcoeff
endif
```

config/builds.csh
```diff
@@ -18,7 +24,8 @@ setenv BuildCompiler 'gnu-openmpi'
 # Note: at this time, all executables should be built in the same environment, one that is
 # consistent with config/environmentForJedi.csh
 
-set commonBuild = /glade/work/guerrett/pandac/build/mpas-bundle_gnu-openmpi_16MAY2022
+# default build directory
+set commonBuild = ${commonBuilddefault}
```
jjguerrette: Here I think you can just use `commonBuild` as the name of the YAML key.  There is no need to add the extra `commonBuilddefault` variable.

Thus,
```
$setLocal commonBuild
```
is more concise.

config/jedi/ObsPlugs/hofx/base/iasi_metop-a.yaml
```diff
@@ -0,0 +1,19 @@
...
...,3440,3442,3444,3446,3448,3450,3452,3454,3458,3467,3476,3484,3491,3497,3499,3504,3506,3509,3518,3527,3555,3575,3577,3580,3582,3586,3589,3599,3610,3626,3638,3646,3653,3658,3661,3673,3689,3700,3710,3726,3763,3814,3841,3888,4032,4059,4068,4082,4095,4160,4234,4257,4411,4498,4520,4552,4567,4608,4646,4698,4808,4849,4920,4939,4947,4967,4991,4996,5015,5028,5056,5128,5130,5144,5170,5178,5183,5188,5191,5368,5371,5379,5381,5383,5397,5399,5401,5403,5405,5446,5455,5472,5480,5483,5485,5492,5497,5502,5507,5509,5517,5528,5558,5697,5714,5749,5766,5785,5798,5799,5801,5817,5833,5834,5836,5849,5851,5852,5865,5869,5881,5884,5897,5900,5916,5932,5948,5963,5968,5978,5988,5992,5994,5997,6003,6008,6023,6026,6039,6053,6056,6067,6071,6082,6085,6098,6112,6126,6135,6140,6149,6154,6158,6161,6168,6174,6182,6187,6205,6209,6213,6317,6339,6342,6366,6381,6391,6489,6962,6966,6970,6975,6977,6982,6985,6987,6989,6991,6993,6995,6997,6999,7000,7004,7008,7013,7016,7021,7024,7027,7029,7032,7038,7043,7046,7049,7069,7072,7076,7081,7084,7089,7099,7209,7222,7231,7235,7247,7267,7269,7284,7389,7419,7423,7424,7426,7428,7431,7436,7444,7475,7549,7584,7665,7666,7831,7836,7853,7865,7885,7888,7912,7950,7972,7980,7995,8007,8015,8055,8078
```
jjguerrette: Does the `HofX` application run in a reasonable time?  If not, it is OK to subset the channels in a similar way to what you have in the `variational` yamls.  Up to you though.

config/jedi/ObsPlugs/variational/base/iasi_metop-a.yaml
```diff
@@ -0,0 +1,22 @@
...
+    name: iasi_metop-a
+    obsdatain:
+      obsfile: {{InDBDir}}/iasi_metop-a_obs_{{thisValidDate}}.h5
+      max frame size: 80000
+    obsdataout:
+      obsfile: {{OutDBDir}}{{MemberDir}}/{{obsPrefix}}_iasi_metop-a.h5
+      max frame size: 80000
+    simulated variables: [brightness_temperature]
+    channels: &iasi_metop-a_channels 16,38,49,51,55,57,59,61,63,66,70,72,74,79,81,83,85,87,104,106,109,111,113,116,119,122,125,128,131,133,135,138,141,144,146,148,151,154,157,159,161,163,167,170,173,176,180,185,187,193,199,205,207,210,212,214,217,219,222,224,226,230,232,236,239,243,246,249,252,254,260,262,265,267,275,282,294,296,299,303,306,323,327,329,335,345,347,350,354,356,360,366,371,373,375,377,379,381,383,386,389,398,401,404,407,410,414,416,426,428,432,434,439,445,457,515,546,552,559,566,571,573,646,662,668,756,867,906,921,1027,1046,1121,1133,1191,1194,1271,1479,1509,1513,1521,1536,1574,1579,1585,1587,1626,1639,1643,1652,1658,1671,1786,1805,1884,1991,2019,2094,2119,2213,2239,2271,2321,2398,2701,2889,2958,2993,3002,3049,3105,3110,5381,5399,5480
+# using a subset of these 616 channels where use_flag = 1
```
jjguerrette: Please add a clarification of why a subset is used.  Perhaps "to reduce computational cost"?  Also say how many channels are used.

config/jedi/ObsPlugs/variational/filtersWithBias/iasi_metop-a.yaml
```diff
@@ -0,0 +1,155 @@
...
+        obserr_dtempf: [0.50,2.00,4.00,2.00,4.00]
+    maxvalue: 1.0e-12
+    action:
+      name: reject
+#  Wavenumber Check (not needed when assimilating a subset channels that does not include any channels listed below)
```
jjguerrette: `a subset channels` --> `a subset of channels`

jjguerrette (commented):
Only minor comments.  Looks good.

PrepJEDI.csh
ibanos90: Changed it.

PrepJEDI.csh
ibanos90: The overall goal is to find previous existing satellite bias correction coefficients. When there are missing observation files, the coefficients are not updated and then we need to search if there are any coefficients updated in previous cycles, otherwise use the initial coefficients created offline. Certainly your suggestion of functionality is much better. I only changed the name of the variable `found` to `foundsatbias`. Thanks!

config/builds.csh
ibanos90: I agree. Thanks for this suggestion.

config/jedi/ObsPlugs/hofx/base/iasi_metop-a.yaml
ibanos90: It makes sense. For consistence, I kept it as in the `variational` yamls

jjguerrette:
@ibanos90, you could add IASI to _one_ of the automated tests to make sure the functionality gets exercised.

config/jedi/ObsPlugs/variational/base/iasi_metop-a.yaml
ibanos90: The comment was updated as suggested.

config/jedi/ObsPlugs/variational/filtersWithBias/iasi_metop-a.yaml
ibanos90: The comment was updated as suggested.

ibanos90:
> @ibanos90, you could add IASI to _one_ of the automated tests to make sure the functionality gets exercised.

Sure! I can do that.

jjguerrette:
> > @ibanos90, you could add IASI to _one_ of the automated tests to make sure the functionality gets exercised.
> 
> Sure! I can do that.

Up to you.  I can approve now if that seems like a lot of work.

liujake:
Considering Ivette will be the only user/tester of this IASI stuff for a while, I suggest not spending time to make scripts generic for other purposes. More important is to obtain results and conclude this task.

jjguerrette:
Oh I meant to just add IASI to the existing 120km 3denvar test, not to add an extra test.

jjguerrette:
> Considering Ivette will be the only user/tester of this IASI stuff for a while, I suggest not spending time to make scripts generic for other purposes. More important is to obtain results and conclude this task.

I would like the functionality to be merged so that an old branch does not get out of date in the next 3 months or longer before we revisit this effort.

ibanos90:
> Oh I meant to just add IASI to the existing 120km 3denvar test, not to add an extra test.

Sorry, I misunderstood. Could you please verify if the changes I committed are what you asked?

scenarios/testinput/3denvar_OIE120km_WarmStart.yaml
```diff
@@ -7,15 +7,44 @@ workflow:
...
+        iasi_metop-b: /glade/p/mmm/parc/ivette/pandac/Observations/iasi/1h
+        iasi_metop-c: /glade/p/mmm/parc/ivette/pandac/Observations/iasi/1h
 experiment:
   ExperimentName: '3denvar_OIE120km_WarmStart_TEST'
 model:
   outerMesh: 120km
   innerMesh: 120km
   ensembleMesh: 120km
+builds:
+  commonBuild: /glade/p/mmm/parc/ivette/pandac/jedi/build22MAY_th
```
jjguerrette: I'm thinking that this line presents a problem, because we want all the tests to run with the same build, possibly an experimental one such as the latest develop branch.  Therefore I think it is best to update the build used for all simulations under `scenarios/base/builds.yaml` as part of this PR instead of adding `builds.commonBuild` to `scenarios/testinput/3denvar_OIE120km_WarmStart.yaml`.

scenarios/testinput/3denvar_OIE120km_WarmStart.yaml
```diff
@@ -7,15 +7,44 @@ workflow:
...
+    120km:
+      120km:
+        3denvar:
+          baseSeconds: 1200
+          memory: 109
 job:
-  CPQueueName: economy
-  NCPQueueName: economy
+  CPQueueName: premium
+  NCPQueueName: premium
```
jjguerrette: Queue names should remain as `economy`

liujake:
Yes. This function should be merged. As I am not using it yet, I am not sure what this actually means. As there are quit a lot scenarios of running JEDI-MPAS with a new function, it could be more work to implement it for all scenarios (I am not sure). I just suggest you make a trade off (for now) about full vs. partial functionality if too much extra work is needed for full function. My sense for the workflow is to provide a good basis for people to adopt/modify it for their own applications, instead of making it a 'click to run' package without knowing it (good but beyond what we can do now).

jjguerrette:
@liujake, Ivette's work is entirely contained in her own scenario.  However, it is also very easy to adopt the IASI functionality to any scenario.  That is the nature of the overall design of the workflow, without adding much extra work.

scenarios/testinput/3denvar_OIE120km_WarmStart.yaml
ibanos90: Then, should we use that build for all simulations?

jjguerrette: Yes, as part of this PR.  We will want to merge the mpas-jedi PR on the same day to ensure that whatever the next build is after yours also includes the tropopause pressure functionality.

liujake:
@jjguerrette Sounds good! It is a bit scary in seeing 43 files touched without looking into them. :-)

scenarios/testinput/3denvar_OIE120km_WarmStart.yaml
```diff
@@ -7,15 +7,44 @@ workflow:
...
+  PANDACArchive:
+    instrumentsAllowingBiasCorrection:
+    - iasi_metop-a
+    - iasi_metop-b
+    - iasi_metop-c
+    IODADirectory:
+      variational:
+        iasi_metop-a: /glade/p/mmm/parc/ivette/pandac/Observations/iasi/1h
+        iasi_metop-b: /glade/p/mmm/parc/ivette/pandac/Observations/iasi/1h
+        iasi_metop-c: /glade/p/mmm/parc/ivette/pandac/Observations/iasi/1h
```
jjguerrette: L10-19 are already in `scenarios/base/observations.yaml`.  So they are not needed here.

scenarios/testinput/3denvar_OIE120km_WarmStart.yaml
ibanos90: That makes sense. I need to check a few things in the mpas-jedi PR yet.

scenarios/base/observations.yaml
```diff
@@ -120,6 +126,11 @@ observations:
         ahi_himawari8: /glade/work/guerrett/pandac/obs/AHIASR/ioda-v2/IODANC_SUPEROB101X101_no-bias-correct
         ahi-clr_himawari8: /glade/work/guerrett/pandac/obs/AHIASR/ioda-v2/IODANC_SUPEROB101X101_no-bias-correct
 
+        ## iasi
+        iasi_metop-a: /glade/p/mmm/parc/ivette/pandac/Observations/iasi/1h
+        iasi_metop-b: /glade/p/mmm/parc/ivette/pandac/Observations/iasi/1h
+        iasi_metop-c: /glade/p/mmm/parc/ivette/pandac/Observations/iasi/1h
```
jjguerrette, line 132: My previous comment was not completely correct.  The `iasi` instruments need to be added to `PANDACArchive.instrumentsAllowingBiasCorrection`.

scenarios/testinput/3denvar_OIE120km_WarmStart.yaml
ibanos90: That's right! I will remove it.

scenarios/base/observations.yaml
ibanos90: I had added them to the scenario, but can move them to the `observations.yaml` as the default option when `biasCorrection: True` and using `PANDACArchive` data.

jjguerrette: We currently do not have bias correction enabled in any test cases, which is probably an oversight.  Although this is already getting more complicated than I expected.  Maybe forget about adding IASI to a test for now.

ibanos90: Ok, I will move changes back and leave the IASI scenario as one of the available scenarios. I will also leave the common build as the one in `develop`, correct?

runs/test.yaml
```diff
@@ -9,6 +9,7 @@ run:
     3dvar_OIE120km_ColdStart,
     3denvar_O30kmIE60km_WarmStart,
     eda_OIE120km_WarmStart,
+    IASI_3denvar_OIE120km_WarmStart,
```
jjguerrette: This line needs to be removed too.  After that, please run tests to make sure they still pass.  Then we merge before I have any more crazy ideas.

scenarios/base/observations.yaml
jjguerrette: Yes, that's good.

runs/test.yaml
ibanos90: Yes, I forgot about it. I removed it in the last commit. Thanks for your review!

ibanos90:
Ran the tests again and all completed successfully.

scenarios/base/hofx.yaml
```diff
@@ -45,6 +48,10 @@ hofx:
 # OPTIONS: False (not enabled yet)
   biasCorrection: False
 
+# method for the tropopause pressure determination used in the
+# cloud detection filter for infrared observations
+  tropprsMethod: thompson
```
jjguerrette, line 54: Please add `# OPTIONS: thompson, wmo` similar to other config elements.

scenarios/base/variational.yaml
```diff
@@ -13,6 +13,10 @@ variational:
 # OPTIONS: True/False
   biasCorrection: False
 
+# method for the tropopause pressure determination used in the
+# cloud detection filter for infrared observations
```
jjguerrette, line 17: Please add `# OPTIONS: thompson, wmo` similar to other config elements.

jjguerrette (commented):
Two more comments requested. Ready to merge after that.

scenarios/base/hofx.yaml
ibanos90: Thanks for this comment! I added that and included that the current build used only works for thompson. This needs to be updated when I have the new build of the code with both methods included.

scenarios/base/variational.yaml
ibanos90: Added it. Thanks!

scenarios/base/hofx.yaml
jjguerrette: OK, thanks.  Let's update the comment in a follow-on PR.

jjguerrette (approved):

ibanos90:
Thanks for all your comments and suggestions @jjguerrette!

jjguerrette:
> @jjguerrette Sounds good! It is a bit scary in seeing 43 files touched without looking into them. :-)

I agree that adding one instrument should not require modifying so many files.  There are some places to improve the workflow that would reduce that count some, but I will wait to do it until we are doing other refactoring.

At least some of the files did not need to be touched in this PR (amsua-related), but the changes are needed eventually anyway.  If the tropopause pressure method was kept fixed instead of making it an option, that would avoid a couple files as well.  The IASI obs required changing how we handle missing channel numbers and also missing bias correction files, which will make radiance assimilation more robust in all our cycling scenarios.  So there were lots of little details that made IASI require special treatments that the workflow was not previously designed to handle.

config/jedi/ObsPlugs/variational/filters/amsua_n15.yaml
```diff
@@ -22,3 +16,9 @@
     minvalue: 1.0e-12
     action:
       name: reject
+    maxvalue: 0
```
jjguerrette, line 19: @ibanos90, this line change introduced a bug for amsua_n15.  The PreQC filter should be like below:
```yaml
  - filter: PreQC
    maxvalue: 0
```

This is why we do not assimilate any `amsua_n15` observations in current experiments.

ibanos90: Thanks for catching this @jjguerrette! I will correct it in a next PR.

PR #111, opened 2022-06-22
Add more forecast diagnostics

jjguerrette:
### Description
Add more forecast diagnostics, consistent with known MPAS-A verification use-cases

### Issue closed

None

### Tests completed


config/mpas/forecast/stream_list.atmosphere.diagnostics
```diff
@@ -10,3 +10,15 @@ cldfrac_tot_UM_max
 cldfrac_tot_UM_rand
 cldht_top_UM
 cldht_base_UM
+height_250hPa
+height_500hPa
```
jjguerrette: Per @junmeiban, only `height_500hPa` is needed for 500hPa geopotential height ACC score.  Others are added for possibly useful verification with minimal extra cost.  That avoids the need to rerun forecasts for those variables should we be interested to consider them.

junmeiban: Thanks JJ. Would you please add "xtime"? Just as Jake's diag files.

jjguerrette: Sorry for missing that.  Should I just copy Jake's entire `stream_list.atmosphere.diagnostics`?

liujake: I think it is better you at least include all variables in my stream file, then can add more variables if you like (e.g., z at 250 hPa).

jjguerrette: Done.

junmeiban (approved):

PR #121, opened 2022-07-12
Independent Model/Obs verification and offline obs/analysis generation

### LINKED ISSUE #116: Independent Model/Obs verification and offline obs/analysis generation
1. Sometimes only one of the model-space or obs-space verification needs to be run.  Provide 2 YAML switches that allows for them to be run independently

2. The observation and GFS analysis processing available in the ColdStart and RealTime runs can be generalized to run for any dates without running Variational, Forecast, or verification tasks.  Add two new scenarios that enable anybody to generate observations or GFS analyses for their time period and mesh.

Work is mostly complete in `feature/2_GenerateAndIndependentModelObs` with base branch `feature/1_modeldata2yaml`.

jjguerrette:
### Description

1. Two new scenarios are added, `GenerateGFSAnalyses` and `GenerateObs`, which use the GFS analysis generation and observation file generation features from the real-time scenario, applicable for any dates of the users choosing that are valid for RDA.  So if we decide to run on a new mesh or during a new time period, it is easy to generate the verifying data without requiring somebody to do it manually.  Then the data can be moved to a secure location for all group members' use in similar experiments.

2. Two new verification options are added, `workflow.VerifyAgainstObservations` and `workflow.VerifyAgainstExternalAnalyses`, which take True or False as values.  By default, both are True, following the current default behavior in develop.  This enables users to select only one verifying data set or the other (obs-space or model-space), which is particularly useful when re-running verification for some reason.

3. `clobber_mode="truncate"` is added to the diagnostics stream during forecasts

4. Update `GenerateABEInflation.csh`

5. Other small refactorings and changes to 60km EDA resources

### Issue closed

Closes #116 

### Tests completed
 - [x] eda_OIE120km_WarmStart


ibanos90 (approved):
Looks good. Ran the tests and all completed successfully.

include/verifyobs.rc
```diff
@@ -81,11 +65,10 @@
       graph = '''
         DataAssimFinished => MeanAnalysis => ExtendedMeanFC
         ExtendedMeanFC => HofXMeanFC
-        ExtendedMeanFC => VerifyModelMeanFC
+        HofXMeanFC:succeed-all => VerifyObsMeanFC
+        VerifyObsMeanFC:succeed-all => CleanHofXMeanFC
   {% for dt in ExtendedFCLengths %}
         PrepareObservations[+PT{dt}H] => HofXMeanFC{{dt}}hr
```
jjguerrette: Note that this line does not work as currently written.  I will remove the dependence on PrepareObservations for extended forecast verification for now as I already have a fix in a follow-on branch.  This dependency is only needed if we ever decide to run extended forecasts while the real-time system is running.  A similar dependency on external analysis preparation is needed in `verifymodel.rc`.

jjguerrette:
Re-ran tests with latest develop merged.  All passed.  Merging.

PR #123, opened 2022-07-12
Enable dual-mesh 3dvar in cold-start scenarios that can use any `outerMesh`

### LINKED ISSUE #122: Enable dual-mesh 3dvar in cold-start scenarios that can use any `outerMesh`
### Problem statement
The Variational application requires a `static.{{nCells}}.nc` and `templateFields.{{nCells}}.nc` file on each mesh (1 per Geometry) that is used in DA, i.e., `outerMesh`, `innerMesh`, and `ensembleMesh`.  Currently we can only do dual-mesh 3dvar or 3denvar on WarmStart runs for which we already have `x1.{{nCells}}.init.*.nc` files pre-generated somewhere on Cheyenne.

Currently for cold-start scenarios, we generate `x1.{{nCellsOuter}}.init.*.nc` as the cold-start initial condition of the first forecast.  This can also be used as the `static.{{nCellsOuter}}.nc` file for any of the Forecast, Variational, HofX, etc... applications.  Then the background state file can be used as `templateFields.{{nCellsOuter}}.nc` for Variational and HofX applications.  For dual-mesh 3DVar, we also need `x1.{{nCellsInner}}.init.*.nc` (plus `x1.{{nCellsEnsemble}}.init.*.nc` for 3denvar).  Although `nCellsInner==nCellsEnsemble` currently, Benjamin Menetrier is working on allowing them to be different.  Likely we could benefit from such a configuration by keeping the inner loop mesh coarser than the ensemble mesh, for example.

### Solution
The easiest way to enable dual-mesh 3dvar (and eventually dual-mesh 3denvar) in cold-start scenarios is to run `init_atmosphere` on both of the inner loop and ensemble meshes using the identical ungrib file that was used on the outer loop mesh.  This entails templating some of the `ExternalAnalyses` generation mechanisms to work with all three meshes, and not only the outer loop mesh.

### Why would we want to be able to do this?
Once enabled, any user can run 3dvar for any date they wish as long as the `innerMesh` is `120km` or `60km` (only two meshes for which we have a static B).  The outer mesh can be freely changed, including to variable resolution meshes, in order to fit particular applications.  After additionally completing #118 and #119, we will be able to start new hybrid 3denvar experiments for any date.

### Working branch
Fully enabled in `feature/dualMeshColdStart`, including a new `3dvar_O30kmIE60km_ColdStart` automated test.

jjguerrette:
### Description
1. Dual-mesh 3dvar is enabled in cold-start scenarios by running the init_atmosphere on both of the inner loop and ensemble meshes using the identical ungrib file that was used on the outer loop mesh. The `ExternalAnalysisToMPAS-{{mesh}}` cylc task has multiple mesh-specific versions, each feeding the appropriate mesh name to `ExternalAnalysisToMPAS.csh`.  The same templating applies for `LinkExternalAnalysis-{{mesh}}`.

2. Any user can run 3dvar for any date they wish as long as the `model.innerMesh` and `model.ensembleMesh` are both equal to either 120km or 60km (only two meshes for which we have a static B). The outer mesh can be freely changed, including to variable resolution meshes, in order to fit particular applications. There is no need to generate any of the inner, outer, or ensemble mesh data offline, except for (1) static B, and (2) ensemble forecasts and (3) localization for 3denvar or 3dhybrid.

3. All substitution variables in `streams.atmosphere` and `streams.init_atmosphere` are enclosed in double braces.

4. `ArgMesh` is also added to `forecast.csh`, which might eventually be used to run ensemble and central deterministic forecast members on different meshes in a single workflow.  That functionality is not enabled here, however.

### Issue closed

Closes #122

### Tests completed
 - [x] eda_OIE120km_WarmStart


ibanos90 (approved):
Thanks for this PR @jjguerrette. This could be used for the RealTime run if needed in the future. 
Ran the tests and all completed successfully.

jjguerrette:
Re-ran tests with latest develop merged. Fixed a bug in `/forecast.csh`.  All tests passed. Merging.

PR #130, opened 2022-07-25
Include gnu version in module load

### LINKED ISSUE #129: Possible bug in environment setup
https://github.com/NCAR/MPAS-Workflow/blob/179724cbfa8a5396cb94b4c631ec3710be1017fa/config/environmentJEDI.csh#L10

The above line may need to be changed to include the module version number, as below
```
setenv BuildCompiler 'gnu-openmpi/10.1.0'
```

Even if this is not a required change, it is a recommended change.

jjguerrette:
### Description
Eliminates a possible source of errors by ensuring we use the correct gnu-openmpi module on Cheyenne.  Nobody reported this is causing errors, but we need to follow common practices between the bundle build and the workflow.

### Issue closed

Closes #129 

### Tests completed

### DIFF
--- a/config/environmentJEDI.csh
+++ b/config/environmentJEDI.csh
@@ -7,7 +7,7 @@ setenv config_environmentJEDI 1
 # {compiler}-{mpi-implementation} combination that selects the JEDI module used to build
 # the executables described in config/builds.csh
 # OPTIONS: gnu-openmpi, intel-impi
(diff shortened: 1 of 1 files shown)


metdyn (approved):
The statement `setenv BuildCompiler 'gnu-openmpi'` appears only once throughout the repo (in this file).  This updated gnu-openmpi version follows mpas-bundle current setting. Agree!

PR #145, opened 2022-08-25
Prepare external analyses past final cycle point

jjguerrette:
### Description

`PrepareExternalAnalyses*` and `PrepareObservations` tasks are modified to include forecast duration offsets between `thisCycleDate` and `thisValidDate`, such that later cycle data can be processed in an earlier cycle.  This allows for extended forecast verification beyond the cylc `final cycle point`.  In this bugfix branch there are redundancies between cycles when preparing observations and external analyses.  For example, the external analysis used to verify the 24 hour forecast from 2018051400 is identical to the one used to verify the 0 hour forecast from 2018051500.  Removing those redundancies will require some smart inter-cycle triggering that checks whether the required data is already present, which can wait for a follow-on PR.  The extended forecast functionality is primarily used for the 2018 test period, such that the redundant tasks are duplicating link creation and not any computationally expensive procedures.

In addition to that bugfix, some flexibility is added for running extended forecasts separately from the obs-space and model-space verification, as requested in #143.  When `VerifyExtendedMeanFC` is `True`, and both `VerifyAgainstObservations` and `VerifyAgainstExternalAnalyses` are `False`, only the extended forecast will be run and not the verification.  The opposite is not true however.  That is to say, setting `VerifyExtendedMeanFC` to `False` (irrespective of the values for `VerifyAgainstObservations` and `VerifyAgainstExternalAnalyses`) will turn off both the extended forecast and the corresponding verification.  So #143 is not completely addressed.

### Issue closed

Closes #144 

### Tests completed
 - [x] 3dvar_OIE120km_WarmStart
 - [x] 3denvar_OIE120km_WarmStart
 - [x] 3dvar_OIE120km_ColdStart
 - [x] 3dvar_O30kmIE60km_ColdStart
 - [x] 3denvar_O30kmIE60km_WarmStart
 - [x] eda_OIE120km_WarmStart

PR #150, opened 2022-08-29
Always use valid date for data retrieval

### LINKED ISSUE #149: Valid date substrings need to be used instead of cycle date substrings
The following scripts were modified in #145 to include a delta time offset from the cycle date:
`GetGFSAnalysisFromFTP.csh`
`GetGFSAnalysisFromRDA.csh`
`GetObs.csh`

However, those scripts use substrings of the cycle date to grab files from external sources with very specific naming conventions.  These scripts need to be updated to use `thisValidDate` substrings instead of `thisCycleDate` substrings for the date-specific naming conventions.

jjguerrette:
### Description
Fixes a bug introduced in #145 that would have impacted extended forecast verification when observations and external analyses are retrieved online in the workflow, such as in real-time applications.  The date strings for file retrieval now use the valid date instead of the cycle date.  This would not have impacted the 2018 historical period applications or background forecast verification (0-hr time offset between valid and cycle dates).

### Issue closed

Closes #149 

### Tests completed
Only ColdStart tests and RealTime application are impacted

### DIFF
--- a/GetGDASanalysis.csh
+++ b/GetGDASanalysis.csh
@@ -8,12 +8,15 @@ date
 source config/model.csh
 source config/experiment.csh
 source config/builds.csh
-set yyyymmdd = `echo ${CYLC_TASK_CYCLE_POINT} | cut -c 1-8`
-set yy = `echo ${CYLC_TASK_CYCLE_POINT} | cut -c 1-4`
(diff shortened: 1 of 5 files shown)


ibanos90 (approved):
Thanks for this PR @jjguerrette. I ran the tests and all completed successfully.

PR #156, opened 2022-09-20
Add scenario to download GDAS analyses in real-time and other small modifications

ibanos90:
### Description
This PR adds a separate scenario for the downloading of GDAS analyses (atm, sfc, and sfluxgrb) in near real-time from the NCEP FTP. It adds the necessary task specification and critical path for this stand-alone run that only downloads those data. Also, in this PR are included small changes in the computational requirements of the scenario to run 3denvar OIE120km experiments using 6h IASI observations. The `Retry` option to search for new observations and GFS analyses was also modified to 10minutes instead of 5minutes in order to decrease the frequency this request is submitted to Casper. 

Additionally, it is added the possibility to use the GFS 6-h forecast from previous cycle as the "GFS 0-h forecast" for current cycle, for model space verification purposes. While running a few days the real-time run using RDA data on Cheyenne, it was found out that for some specific days and synoptic hours there is missing the GFS 0-h forecast that we use for the model space verification of MPAS 6-h forecast. Therefore, this will only be of relevance for retrospective experiments using raw data. This was solved by adding an if statement in the script  `GetGFSAnalysisFromRDA.csh` which will allow us to use the previous 6-h forecast.

### Tests completed
 - [x] eda_OIE120km_WarmStart

### DIFF
--- a/GetGFSAnalysisFromRDA.csh
+++ b/GetGFSAnalysisFromRDA.csh
@@ -26,6 +26,7 @@ set yy = `echo ${CYLC_TASK_CYCLE_POINT} | cut -c 1-4`
 set hh = `echo ${CYLC_TASK_CYCLE_POINT} | cut -c 10-11`
 set thisCycleDate = ${yymmdd}${hh}
 set thisValidDate = `$advanceCYMDH ${thisCycleDate} ${ArgDT}`
+set prevValidDate = `$advanceCYMDH ${thisCycleDate} -6`
 
 set res = 0p25
 set fhour = 000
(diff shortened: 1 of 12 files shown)


jjguerrette (approved):
If we had a full set of tasks going from GDAS analysis bufr to MPAS x1.init.nc file, then I would prefer to have GDAS added as another kind of `externalanalyses` with its own set of `PrepareExternalAnalysisTasks` and `Vtable`.  However, since such a requirement is not needed in the short term, this solution works.  Good compromise!

ibanos90:
> If we had a full set of tasks going from GDAS analysis bufr to MPAS x1.init.nc file, then I would prefer to have GDAS added as another kind of `externalanalyses` with its own set of `PrepareExternalAnalysisTasks` and `Vtable`. However, since such a requirement is not needed in the short term, this solution works. Good compromise!

That would be great! We can certainly try that in the future!

PR #177, opened 2022-12-22
Remove obsolete 'verbosity' key in the saber section

### LINKED ISSUE #173: bump yamls need update for a key
'verbosity' key no longer exists with the latest saber code. It needs to be removed or commented out, or run will crash.

liujake:
### Description

Comment out 'verbosity' key in the saber section, which does not exist anymore.

### Issue closed
Closes #173

### Tests completed
 
Mostly not needed.  Will test more in follow-on PRs

### DIFF
--- a/config/jedi/applications/3denvar-specific_multivariate.yaml
+++ b/config/jedi/applications/3denvar-specific_multivariate.yaml
@@ -60,7 +60,6 @@ cost function:
           #io_values: [dynamic,dynamic,dynamic,dynamic,dynamic,cloud,cloud,cloud,cloud,cloud]
           io_values: [dynamic,cloud,dynamic,dynamic,dynamic,cloud,cloud,cloud,cloud,cloud]
           lev2d: last
-#          verbosity: main
 {{EnsemblePbMembers}}
 {{EnsemblePbInflation}}
   observations:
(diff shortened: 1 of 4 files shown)


jjguerrette:
Thank you @liujake.  The obsolete lines can simply be removed.  No need to keep as comments or run all the tests.

liujake:
Ok. removed.

jjguerrette:
One question is how this removal will affect the 25OCT2022 build that is still used by default.  Anybody who wants to use the 25OCT2022 build can use the tagged pre-release.  You could update the build too.  Then we might want all tests to complete.  I'm not sure if other functionality is broken in the recent develop build besides EDA.  The eda test will fail with develop.

liujake:
This was found when I compare memory usage between 25Oct and 05Dec code. I am not sure the best way to manage the 'release' of workflow. We do not have to merge this PR now if others still use the older version of code.

jjguerrette:
Actually, I was trying to say it's completely fine to update to a new build from you (as long as it is single precision).  Anybody who is still using 25OCT2022 can continue to do so on their own experiment branches or by checking out the tagged pre-release 0.0.0.  We need not be resistant to using newer builds in the develop branch as long as we have tagged releases to go back to and a documentation of the bundle branches used in the workflow release build.  Eventually the bundle and workflow can get in sync with their releases, but maybe not until the next formal bundle release.

jjguerrette (approved):

PR #187, opened 2023-01-26
Update BUMP yamls

### LINKED ISSUE #182: Update BUMP yamls
See https://github.com/JCSDA-internal/mpas-jedi/pull/822 and https://github.com/JCSDA-internal/saber/pull/424.

byoung-joo:
### Description
* Companion to https://github.com/JCSDA-internal/mpas-jedi/pull/822.  YAML formats are changed for `multivariate strategy` key of BUMP.  No change in the result is expected.
* Build is updated to 25JAN2023.
* StaticB files are regenerated, due to a change in the group name formatting in some BUMP NetCDF files.
We can still use the previous versions of BUMP localization files.

### Issue closed

Closes #182 

### Tests completed
 - [x] eda_OIE120km_WarmStart

### DIFF
--- a/config/jedi/applications/3denvar.yaml
+++ b/config/jedi/applications/3denvar.yaml
@@ -55,7 +55,7 @@ cost function:
             data directory: {{bumpLocDir}}
             files prefix: {{bumpLocPrefix}}
           drivers:
-            multivariate strategy: common
+            multivariate strategy: duplicated
             read local nicas: true
           model:
             level for 2d variables: last
--- a/scenarios/base/variational.yaml
+++ b/scenarios/base/variational.yaml
@@ -247,13 +247,13 @@ variational:
       bumpCovStdDevFile: None
       bumpCovVBalDir: None
     60km:
-      bumpCovDir: /glade/p/mmm/parc/bjung/pandac_common/staticB/20230106_prebuilt/60km.NICAS_00
-      bumpCovStdDevFile: /glade/p/mmm/parc/bjung/pandac_common/staticB/20230106_prebuilt/60km.CMAT_00/mpas.stddev_0p33.2018-04-15_00.00.00.nc
-      bumpCovVBalDir: /glade/p/mmm/parc/bjung/pandac_common/staticB/20230106_prebuilt/60km.VBAL_00
+      bumpCovDir: /glade/p/mmm/parc/bjung/pandac_common/staticB/20230125_prebuilt/60km.NICAS_00
+      bumpCovStdDevFile: /glade/p/mmm/parc/bjung/pandac_common/staticB/20230125_prebuilt/60km.CMAT_00/mpas.stddev_0p33.2018-04-15_00.00.00.nc
+      bumpCovVBalDir: /glade/p/mmm/parc/bjung/pandac_common/staticB/20230125_prebuilt/60km.VBAL_00
     120km:
-      bumpCovDir: /glade/p/mmm/parc/bjung/pandac_common/staticB/20230106_prebuilt/NICAS_00
-      bumpCovStdDevFile: /glade/p/mmm/parc/bjung/pandac_common/staticB/20230106_prebuilt/CMAT_00/mpas.stddev_0p33.2018-04-15_00.00.00.nc
-      bumpCovVBalDir: /glade/p/mmm/parc/bjung/pandac_common/staticB/20230106_prebuilt/VBAL_00
+      bumpCovDir: /glade/p/mmm/parc/bjung/pandac_common/staticB/20230125_prebuilt/NICAS_00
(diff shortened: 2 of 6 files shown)


byoung-joo:
Six tests went through without an issue on Cheyenne. I have compared several results with previous version of MPAS-Workflow, but I don't have the results for all scenarios. The test results are located in `/glade/scratch/bjung/pandac/*_TEST.25JAN2023`.

Also, is the configuration `config/jedi/applications/3denvar-crossed.yaml` used in MPAS-Workflow ? 
I just changed the `multivariate strategy` key, but this was not tested and might not work properly...

jjguerrette:
> Six tests went through without an issue on Cheyenne. I have compared several results with previous version of MPAS-Workflow, but I don't have the results for all scenarios. The test results are located in `/glade/scratch/bjung/pandac/*_TEST.25JAN2023`.

As long as the tests completed, as in the cylc suites complete and exit successfully, that is enough.  I do not check the results, because we do not have references.  More work to be done there.

> Also, is the configuration `config/jedi/applications/3denvar-crossed.yaml` used in MPAS-Workflow ? I just changed the `multivariate strategy` key, but this was not tested and might not work properly...

That yaml is not used in the tests or by any example scenarios.  It was a placeholder for when we return to the all-sky work.  Whoever gets back to that work can open a new issue/PR to fix the yaml if needed.

config/jedi/applications/3denvar.yaml
```diff
@@ -55,7 +55,7 @@ cost function:
             data directory: {{bumpLocDir}}
             files prefix: {{bumpLocPrefix}}
           drivers:
-            multivariate strategy: common
+            multivariate strategy: duplicated
```
jjguerrette, line 58: Does `common` need to be changed to `duplicated` in the localization generation too?  Asking so I can update my own offline yamls.

scenarios/base/variational.yaml
```diff
@@ -247,13 +247,13 @@ variational:
       bumpCovStdDevFile: None
       bumpCovVBalDir: None
     60km:
-      bumpCovDir: /glade/p/mmm/parc/bjung/pandac_common/staticB/20230106_prebuilt/60km.NICAS_00
-      bumpCovStdDevFile: /glade/p/mmm/parc/bjung/pandac_common/staticB/20230106_prebuilt/60km.CMAT_00/mpas.stddev_0p33.2018-04-15_00.00.00.nc
-      bumpCovVBalDir: /glade/p/mmm/parc/bjung/pandac_common/staticB/20230106_prebuilt/60km.VBAL_00
+      bumpCovDir: /glade/p/mmm/parc/bjung/pandac_common/staticB/20230125_prebuilt/60km.NICAS_00
```
jjguerrette, line 250: Should I update localization files too?

jjguerrette (commented):
Thank you BJ. Just a couple questions.

config/jedi/applications/3denvar.yaml
byoung-joo: Yes, it should be updated there!

scenarios/base/variational.yaml
byoung-joo: There is no change in the localization files (specifically for group name `common`), so we can use the existing ones.

byoung-joo: It is a bit complex... the `multivariate strategy` key is changed from `common` to `duplicated`, BUT the NetCDF group name `common` stays the same as before in BUMP NetCDF file.

jjguerrette: OK, then I will not update them.  Do the yaml settings below look correct to you for generating localization?

```yaml
  drivers:
    ## always on
    compute nicas: true
    write local nicas: true
    write global nicas: true
    write nicas grids: true

    ## same length scales for all variables
    multivariate strategy: duplicated
  nicas:
    resolution: 8.0
    explicit length-scales: true
    ## duplicated
    horizontal length-scale:
      - variables: [duplicated]
        value: HLocLengthKMe3
    vertical length-scale:
      - variables: [duplicated]
        value: VLocLengthKMe3
```

byoung-joo:
> As long as the tests completed, as in the cylc suites complete and exit successfully, that is enough. I do not check the results, because we do not have references. More work to be done there.

Yes, all cylc suites complete and exit successfully!

> That yaml is not used in the tests or by any example scenarios. It was a placeholder for when we return to the all-sky work. Whoever gets back to that work can open a new issue/PR to fix the yaml if needed.

Thank you for your explanation, @jjguerrette. We can continue this part latertime! perhaps when we update the build version for benchmark experiment!

scenarios/base/variational.yaml
byoung-joo: Let me double check this !

jjguerrette: I'm checking right now.  Will report back.

jjguerrette (approved):

scenarios/base/variational.yaml
jjguerrette: I changed `variables: [duplicated]` to `groups: [common]`.  That appears to be working.

Issue #194, opened 2023-01-31
Port yaml parsing to python


jjguerrette:
Two motivations.  The current parsing of yamls in tcsh is:
1. incredibly slow with lots of redundant reads of entire scenario yaml files
2. difficult to maintain, extend, and document
 
Mostly complete in two feature branches:
1. `feature/0_pythonConfigParse`
4. `feature/1_pythonConfigParse`

The first branch makes the transition to python feasible by dividing `drive.csh` into 4 different named suites (`Cycle`, `ForecastFromExternalAnalyses`, `GenerateExternalAnalyses`, and `GenerateObs`), and isolating the suite aspects and configuration aspects associated with each application (e.g., `variational`, `forecast`, `initic`, etc...) into a single shell script.

The second branch converts those non-overlapping functionalities (multiple suites and multiple applications) into python classes.  It also moves all YAML parsing to python, which is at least an order of magnitude faster than the current parsing in shell scripts.

There is still room for later improvement, but these feature branches together unties a lot of complex inter-dependencies that exist in the current shell scripts.  Eventually this functionality can be leveraged to improve YAML design, because python can parse any valid YAML structure into a hierarchy of dictionaries.  These branches have been sitting around for a while.  I delayed merging because there will be some interruption to individuals' experiment YAMLS.  I would like to merge them soon, because merging develop takes a lot of work, especially when people add new YAML options.

Please reach out to discuss further.

PR #199, opened 2023-02-23
First step to python-ized scenario configuration parsing

jjguerrette:
### Description
Major refactoring.  Over 100 files touched.  Divides `drive.csh` into 4 different named suites (Cycle, ForecastFromExternalAnalyses, GenerateExternalAnalyses, and GenerateObs). One benefit of this PR is that some settings that are only needed for the Cycle suite can be removed for other suites (e.g., see scenarios/GenerateObs.csh).

Isolates the suite aspects and configuration aspects associated with independent applications (e.g., variational, forecast, initic, etc...) into their own shell script (i.e., variational.csh, forecast.csh, initic.csh, etc...).  A follow-on PR will take that design one step further to collect the configuration for each application into a python class.

Some of the cylc suite task descriptions are now automatically generated in those application-specific scripts.  Other tasks (like `HofX*`) are not yet automated, but will be in later work after the cylc dependency generation is also automated.  The automatic dependency generation is partially implemented in `feature/1_pythonConfigParse`, because that is much easier to do in python than in shell scripts.

Note that many scenario yamls have been changed to organize their root keys alphabetically.  This is only for ease of maintenance and is not a requirement for any users.

The number of members is now controlled with `members.n` (defaults to 1) instead of under the `variational` yaml section.  This eases use of the configuration element that controls the number of members across applications (i.e., HofX, Forecast, RTPP) as needed.

These three extended forecast settings are moved from the `workflow` to the `forecast` section:
```
  ExtendedMeanFCTimes: T00
  ExtendedFCLengthHR: 240
  ExtendedFCOutIntervalHR: 24
```

### Issue closed

First part of #194

### Tests completed
 - [x] eda_OIE120km_WarmStart


jjguerrette:
Since this is an intermediate step that merely facilitates the second step, I recommend not getting too attached to the shell script design in the `config` directory.  The application-specific and data resource-specific shell scripts will be replaced by python classes in the follow-on PR.  This PR's main purpose is to untie circular/complex dependencies.  The follow-on PR and later work will further that endeavor.

ibanos90 (approved):
Hi @jjguerrette, thanks for this PR! It is a lot work! I ran the test and all completed successfully.

junmeiban (approved):
Tested and all passed. Thank you so much @jjguerrette !

scenarios/ForecastFromGFSAnalyses.yaml
```diff
@@ -1,31 +1,46 @@
...
-model:
-  outerMesh: 60km
-#TODO: make it so that innerMesh and ensembleMesh are not needed
-  innerMesh: 60km
-  ensembleMesh: 60km
+
+forecast:
+  ExtendedMeanFCTimes: T00,T12
+  ExtendedFCLengthHR: 24
+  ExtendedFCOutIntervalHR: 12
```
jjguerrette, line 11: I want to make a note to all users.  These three extended forecast settings are moved from `workflow` to `forecast`.  There will be additional yaml format changes in the next PR, which will be demonstrated in the example scenario yamls.

scenarios/GenerateObs.yaml
```diff
@@ -22,15 +15,8 @@ experiment:
...
-  resource: "ForecastFromAnalysis"
-externalanalyses:
-  resource: "GFS.RDA"
-model:
-  outerMesh: 120km
-  innerMesh: 120km
-  ensembleMesh: 120km
-variational:
-  DAType: 3dvar
-  nInnerIterations: [0,]
```
jjguerrette, line 36: Another benefit of this PR is that some settings that are only needed for the `Cycle` suite can be removed for other suites (e.g., these ones for `GenerateObs`).

liujake (approved):
Not able to check 100+ files! Fine with me to merge.

PR #251, opened 2023-06-28
Fixed a failure of a test1 case, update verification obs input, and use bending angle

liujake:
### Description
1. The default is set to 36 cores per node for the forecast step, which caused a failure for one 'test1' scenario pointing to Duda's folder missing the MPAS partition file for 288 cores. Set to 32 cores (thus use 256 cores) in the scenario yaml fixed the problem.

2. Update the hofx obs input location for cycling experiments using raw data with bias correction. Those hofx input obs are used for the forecast verification purpose and we'd better use GSI-ncdiag files and bias-corrected radiance data so that the consistent number of obs is used when comparing forecasts from multiple cycling experiments.

3. Use GNSSRO bending angle instead of refractivity

### List of modified files
M       scenarios/3dvar_O30kmIE60km_ColdStart.yaml
M       test/testinput/3dvar_O30kmIE60km_ColdStart.yaml
M       scenarios/defaults/observations.yaml
M       initialize/data/Observations.py

### DIFF
--- a/initialize/data/Observations.py
+++ b/initialize/data/Observations.py
@@ -21,8 +21,8 @@ from initialize.framework.HPC import HPC
 benchmarkObservations = [
   # anchor
   'aircraft',
-  'gnssrorefncep',
-#  'gnssrobndropp1d',
+#  'gnssrorefncep',
+  'gnssrobndropp1d',
   'satwind',
   'satwnd',
   'sfc',
--- a/scenarios/3dvar_O30kmIE60km_ColdStart.yaml
+++ b/scenarios/3dvar_O30kmIE60km_ColdStart.yaml
@@ -22,6 +22,12 @@ model:
 externalanalyses:
   resource: "GFS.RDA"
 forecast:
+  job:
+    30km:
+      nodes: 8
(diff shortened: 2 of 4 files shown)


liujake:
@ibanos90 Bending angle results look good. So I changed the default to use bending angle now. Please review.

junmeiban (approved):

ibanos90:
> @ibanos90 Bending angle results look good. So I changed the default to use bending angle now. Please review.

Sorry that I missed this. These changes look good!

byoung-joo:
@liujake , I noticed that this PR is also merged to `release/2.0.0` branch. I wonder if current `release/2.0.0` branch of MPAS-Workflow is good for running some experiment. Or are there any further modification planned for `release/2.0.0` ?

liujake:
2.0.0 was cut off from the develop Yesterday. So they are the same now. Good to use 2.0.0 branch.

byoung-joo:
> 2.0.0 was cut off from the develop Yesterday. So they are the same now. Good to use 2.0.0 branch.

Thank you for confirming!

PR #257, opened 2023-09-09
Add Cloud_Fraction in cloudyCRTMObsOperator

### LINKED ISSUE #256: Cloud_Fraction
**Describe the bug/question ???**
Cloud_Fraction should be added in cloudyCRTMObsOperator

junmeiban:
### Description
Cloud_Fraction setting should be added in cloudyCRTMObsOperator section. 

### Issue closed
Closes https://github.com/NCAR/MPAS-Workflow/issues/256

### Modified file:
M config/jedi/ObsPlugs/variational/ObsAnchors.yaml

### DIFF
--- a/config/jedi/ObsPlugs/variational/ObsAnchors.yaml
+++ b/config/jedi/ObsPlugs/variational/ObsAnchors.yaml
@@ -34,6 +34,7 @@ _cloudy crtm: &cloudyCRTMObsOperator
   SurfaceWindGeoVars: uv
   Absorbers: [H2O, O3]
   Clouds: [Water, Ice, Rain, Snow, Graupel]
+  Cloud_Fraction: 1.0
   linear obs operator:
     Absorbers: [H2O]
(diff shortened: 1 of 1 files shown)


byoung-joo:
@junmeiban, I don't know this is a right way. Let me think further and investigate my original issue (zero BT) further...

junmeiban:
> @junmeiban, I don't know this is a right way. Let me think further and investigate my original issue (zero BT) further...

Thanks @BJ. I will also investigate it in my case.

liujake:
@junmeiban @byoung-joo I do not think this is right. If you set cloud_fraction = 1 in yaml, code will assign cloud fraction = 1 everywhere. If it is not set in yaml, cloud fraction values (vertical profiles) will be read in from MPAS model diagnosed values interpolated into obs locations, which is what we are doing now (correct way).

junmeiban:
Hi @liujake, I am just thinking of close this draft PR.

Issue #270, opened 2023-12-14
Cleanup for Cylc8 is incorrectly using the Cylc7 mechanism.


jim-p-w:
The cylc cleanup in submit.csh is cleaning up by removing the cylc-run/MPAS-Workflow/<job> directory.
That is causing subsequent failures when rerunning the same job if the cylc-run is configured to link elsewhere (e.g. a users scratch directory).

The proper way to clean up is to issue a `cylc clean` directive.

**To Reproduce**
1. configure cylc to save cylc run files in your scratch directory:
In ~/.cylc/flow/global.cylc add the following:
```
[install]
   [[symlink dirs]]
      [[[localhost]]]
        run = /glade/derecho/scratch/$USER/
```
2. run a cylc workflow:
`./Run.py test/testinput/3denvar_OIE120km_IAU_WarmStart.yaml`
3. rerun the same workflow. The initial workflow will be stopped, but the workflow won't get restarted successfully.

Specifically, the error you get is as follows:
```
./submit.csh rm -r /glade/u/home/jwittig/cylc-run/MPAS-Workflow/jwittig_3dvar_OIE120km_WarmStart_TEST
./submit.csh cylc install --run-name=jwittig_3dvar_OIE120km_WarmStart_TEST
WorkflowFilesError: Symlink dir target already exists: (/glade/u/home/jwittig/cylc-run/MPAS-Workflow/jwittig_3dvar_OIE120km_WarmStart_TEST ->) /glade/derecho/scratch/jwittig/cylc-run/MPAS-Workflow/jwittig_3dvar_OIE120km_WarmStart_TEST
Tip: in future, use 'cylc clean' instead of manually deleting workflow run dirs.
```

PR #271, opened 2023-12-14
Fix the cylc workflow cleanup to use the Cylc8 "cylc clean" command

jim-p-w:
### Description
 * use `cylc clean` when cleaning up before running a new workflow.
     - this is required if using symlinks to put cylc-run data somewhere other than $HOME.
 * update the README w/ additional iniitalizing step when first using symlinks for cylc run data.

### Issue closed
   Fixes Issue [270](https://github.com/NCAR/MPAS-Workflow/issues/270).

 Changes to be committed:
	modified:   README.md
	modified:   submit.csh)

#### Tier 1:
 - [x] eda_OIE120km_WarmStart

#### Tier 2 (optional):
 - [ ] GenerateGFSAnalyses
 - [ ] GenerateObs

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,9 +52,10 @@ It is required to set the content of $HOME/.cylc/flow/global.cylc as follows:
 ```
 The [[pbs_cluster]] entries tell cylc how to submit jobs.
 The [instal] section will create both $HOME/cylc-run/MPAS-Workflow and 
-/glade/derecho/scratch/$USER/cylc-run/MPAS-Workflow directories to be created.
+/glade/derecho/scratch/$USER/cylc-run/MPAS-Workflow directories.
 A symlink will be created in $HOME/cylc-run/MPAS-Workflow for each workflow,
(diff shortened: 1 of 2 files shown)


mos3r3n:
I tested this command before, but it was slower than directly using "rm -r". Meanwhile, using cylc clean would be safer, in my opniion.

jim-p-w:
To elaborate, removing the run directory fails if you are configured to put the run directory somewhere other that $HOME.
After the `rm`, the `cylc install` command failed with:
```
./submit.csh cylc install --run-name=jwittig_3dvar_OIE120km_WarmStart_TEST
WorkflowFilesError: Symlink dir target already exists...
Tip: in future, use 'cylc clean' instead of manually deleting workflow run dirs.
```
And the workflow doesn't get installed and cannot be run.

mos3r3n:
I agree with you. This PR is good.

@mos3r3n Thanks for taking a look Tao!

mos3r3n (approved):

PR #277, opened 2024-01-09
Resource requirements for hofx.

junmeiban:
### Description
This PR is for resource requirements modification for hofx on Derecho.

### Files modified:
M  scenarios/defaults/hofx.yaml

### Tests completed
Ran hofx.

### DIFF
--- a/scenarios/defaults/hofx.yaml
+++ b/scenarios/defaults/hofx.yaml
@@ -3,9 +3,9 @@ hofx:
   job:
     defaults:
       seconds: 1000
-      nodes: 4
-      PEPerNode: 32
-      memory: 109GB
+      nodes: 2
+      PEPerNode: 128
+      memory: 235GB
 
       # cylc retry string
       retry: '0*PT30S'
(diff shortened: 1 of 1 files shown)


mos3r3n (approved):

byoung-joo (approved):
Looks reasonable to me!

scenarios/defaults/hofx.yaml
```diff
@@ -18,15 +18,15 @@ hofx:
     30km:
       seconds: 600
       nodes: 2
```
jim-p-w, line 20: Will running hofx with a 30km resolution fit on one node?

RGNystrom (approved):

scenarios/defaults/hofx.yaml
junmeiban: Oh It depends. If I run hofx by using conventional obs, AMSU-A and MHS, only one node is okay.
 
For my test case (by using amsr2 and gmi), If I use one node, it just stopped without any error information. If I use two nodes, it run successfully. 
See results with one node (amsr2 and gmi) : 
/glade/derecho/scratch/jban/pandac/jban_3dhybrid_amsuaAll/Verification_1node/bg/mean/2018041500/jedi.log.all
See results with two nodes (amsr2 and gmi): 
/glade/derecho/scratch/jban/pandac/jban_3dhybrid_amsuaAll/Verification_2nodes/bg/mean/2018041500/jedi.log.all

If no opposition, I would like to use 2 nodes setting for 30km resolution.

jim-p-w: Thanks for the info Junmei.

jim-p-w (approved):

scenarios/defaults/hofx.yaml
ibanos90: Thanks for the further information @junmeiban! I am using one node so far and had no issues. I wonder if we should keep 1 as the default and then changed it in our experiments according to our needs.

PR #285, opened 2024-02-14
Adjusts Forecast Output for 3D and 4D Applications

### LINKED ISSUE #280: Forecast step in 3D DA should NOT output HOURLY forecast files
Hi @RGNystrom 
When I tried 120km 3DEnVar Cycling, I noticed the forecast step runs 9-hour forecast with hourly output. I believe this is the cause of a failure I encountered. See the PR I just opened. I guess slow disk I/O (this happens often recently) made the forecast job failed. I guess this is due to your modification for 4DEnVar. Can you fix this? e.g., we should only do 9-hour forecast with hourly output for 4DEnVar mode,  but not for any 3D DA mode. I am not sure where to modify to make this configurable during the run-time.

Another thing is that the forecast output is from hour-1 instead of hour-0. We might want to go back to output from hour-0, as hour-0 forecast includes re-diagnosed pressure field, which might not be exactly the same as the analysis file, and could be useful for diagnostic purposes.

RGNystrom:
### Description
Added `FourD` yaml forecast option for 4D applications. When set to `True` the forecast output will be the subwindow frequency and when `False`, the default, the forecast output will be the window length.

The initial forecast output is also now retained instead of deleted.

### Issue closed

Closes #280

#### Tier 1:
 - [x] 4denvar_OIE120km_WarmStart_TEST
 - [x] eda_OIE120km_WarmStart

#### Tier 2 (optional):
 - [ ] GenerateGFSAnalyses
 - [ ] GenerateObs

### DIFF
--- a/initialize/applications/Forecast.py
+++ b/initialize/applications/Forecast.py
@@ -45,6 +45,10 @@ class Forecast(Component):
     # whether to use incremental analysis update
     'IAU': [False, bool],
 
+    ## 4D
+    # whether to use 4D forecast outputs
+    'FourD': [False, bool],
+
     ## post
     # list of tasks for Post
     'post': [['verifyobs', 'verifymodel'], list]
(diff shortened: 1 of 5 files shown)


liujake (approved):
Looks good to me.

PR #298, opened 2024-03-19
Fix the issue that first background files are in double precession while mpas-bundle in single-precession.

mos3r3n:
### Description
Our mpas-bundle is commonly built in single precession. However, the first background files are in double precession (i.e. ensemble background at 00Z 15 April 2018). In that case, the analysis output will report error information "ERROR: MPAS IO Error: Inconsistent redefinition of dimension" (@RGNystrom reported). To solve this issue, the NCO is employed to convert double (float) variables to float (double) variables when the mpas-bundle is built in single (double) precision.

#### Tier 1:

### DIFF
--- a/bin/LinkWarmStartBackgrounds.csh
+++ b/bin/LinkWarmStartBackgrounds.csh
@@ -28,23 +28,24 @@ while ( $member <= $nMembers )
       # Outer loop mesh
       set fcFile = $CyclingFCDirs[$member]/${FCFilePrefix}.${nextFirstFileDate}.nc
       set InitialMemberFC = "$firstbackground__directoryOuter"`${memberDir} 2 $member "${firstbackground__memberFormatOuter}"`
-      ln -sfv ${InitialMemberFC}/${firstbackground__filePrefixOuter}.*.nc ${fcFile}${OrigFileSuffix}
-      set varinfo = `ncdump -h ${fcFile}${OrigFileSuffix} | grep surface_pressure`
-      if ( "$varinfo" =~ "*double*" ) then
-         if ( "$bundlePrecision" == "single" ) then
-            ncpdq -5 --pck_map=dbl_flt ${fcFile}${OrigFileSuffix} ${fcFile}
-         else
-            mv ${fcFile}${OrigFileSuffix} ${fcFile}
-         endif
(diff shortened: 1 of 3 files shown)


bin/LinkWarmStartBackgrounds.csh
```diff
@@ -28,8 +29,23 @@ while ( $member <= $nMembers )
...
       ln -sfv ${InitialMemberFC}/${firstbackground__filePrefixOuter}.*.nc $CyclingFCDirs[$member]/.
-      # rm ${fcFile}
-      cp ${fcFile}${OrigFileSuffix} ${fcFile}
+      # Handling double/single precision files
+      foreach file ($CyclingFCDirs[$member]/*.nc)
+        set varinfo = `ncdump -h ${file} | grep surface_pressure`
+        if ( "$varinfo" =~ "*double*" ) then
+           if ( "$bundlePrecision" == "single" ) then
+              mv ${file} ${file}${OrigFileSuffix}
+              ncpdq -5 --pck_map=dbl_flt ${file}${OrigFileSuffix} ${file}
```
jim-p-w, line 38: Where can I find `ncpdq`? It's not in my path

mos3r3n: It's a function of the nco. You can see I add "module load nco" in suiteBase.py.

RGNystrom (approved):

jim-p-w:
Instead of converting the files on the fly every time this step is run would it be more efficient to convert the files once, and then use the appropriate files based on `bundlePrecision`?

liujake:
@jim-p-w has a good point. It is better to simply convert files offline once, then use them in the workflow, e.g., replace the old double-precision file location with the new single-precision file location.

mos3r3n:
That's right. Since we now use single-precision mpas-bundle only, it would be better that we prepare these files in single-precession rather than converting them when running workflow. @jim-p-w @liujake

liujake:
@mos3r3n You can still use this PR, but fixing this in a different way: provide new location of the first cycle ensemble background files.

liujake:
I think this one can be closed. @mos3r3n . We should have addressed this by converting double precision first background to single precision outside the workflow.

PR #303, opened 2024-04-29
Refactor Clean and Init tasks, move linking jobs to run in the background

ibanos90:
### Description
This PR refactors some of the tasks. The goal is to reduce the number of jobs to be submitted, hence, computing and waiting time for the each cycle. `Clean` tasks are removed and the content is moved to their associated application tasks (e.g., `CleanHofX` content is moved to `HofX.csh`). `InitVariationals.csh` and `InitEnKF.csh` scripts for the YAML and model stage preparation stage are also removed and their content is moved to `PrepJEDI.csh` (only one preparation task is submitted). For RTPP, `Clean` and `Init` tasks and both removed and merged with `RTPP.csh`. Here, the `LinkExternalAnalysis` and `LinkWarmStartBackgrounds` tasks are set to run in the background instead of Casper. These tasks use very little memory and wall-time, then running them in the background should be fine (but let's monitor that).

### Issue closed
None

#### Tier 1:
 - [x] eda_OIE120km_WarmStart

#### Tier 2:
- [x] 3dhybrid-allsky_O30kmIE60km_SpecifiedEnsemble_VarBC


mos3r3n:
good job! @ibanos90

ibanos90:
> good job! @ibanos90

Thanks @mos3r3n!

junmeiban (approved):
Yes. It's faster than before. Thank you so much @ibanos90 !

junmeiban (approved):

junmeiban:
Now it works for scenarios/3dhybrid-allsky_O30kmIE60km_SpecifiedEnsemble_VarBC.yaml. Many thanks @ibanos90 !

ibanos90:
> Now it works for scenarios/3dhybrid-allsky_O30kmIE60km_SpecifiedEnsemble_VarBC.yaml. Many thanks @ibanos90 !

Thanks a lot for the thorough review @junmeiban!

bin/Variational.csh
```diff
@@ -109,6 +111,30 @@ if ( $status != 0 ) then
   exit 1
 endif
 
+# ================================================================================================
+
+# Remove obs-database output files
+# ================================
+if ("$retainObsFeedback" != True) then
+  set member = 1
+  while ( $member <= ${nMembers} )
```
jim-p-w, line 120: I think this looks incorrect. I'm seeing Variational[1-5] fail intermittently, while it is trying to write output files.
I see this fail most often when running `test/testinput/eda_OIE120km_WarmStart.yaml`

When this is running for any given member, it will remove the output for all the other members, so output files for other members can be deleted while they are being written.

Instead of looping through all of the members in a loop it should only remove files for it's own member number, `ArgMember`

ibanos90: Hi @jim-p-w, thanks for reporting this issue. When I tested it for this PR it was working correctly, but I will take a look.

ibanos90: Indeed, the task will fail for the eda task in the case that the jobs end at different times, the files should be removed for each member independently without the loop. Thanks again for reporting this!

jim-p-w: See PR https://github.com/NCAR/MPAS-Workflow/pull/310 for fixes to Variational.csh and EnKF.csh
See issue https://github.com/NCAR/MPAS-Workflow/issues/309 for what needs to be done for EnsembleOfVariational.csh (I haven't yet gotten a working test that runs EnsembleOfVariational.csh)

PR #304, opened 2024-04-29
Add variable mesh 60-3km for forecast (outer)

ibanos90:
### Description
This PR adds specification for the variable mesh 60-3km. The variable `meshRatio` is introduced in order to get correct static, graph and info files for each application in a more generic way. The namelist is also made more configurable, adding high flexibility for the physical parameterizations selection. For this mesh, Thompson microphysics is used for the forecasts, then the `IRVISlandCoeff` is made configurable in order to use the `IGBP` coefficients for the HofX application (default is `USGS`) and the table for Thompson microphysics are added. `ForecastFromGFSAnalysesVariableMesh` scenario is added for this functionality.

At this moment, this mesh does not work fully for data assimilation (dual-resolution). Although many of the changes introduced here work for the DA, some additional modifications are still needed, which will be the subject of a future PR.

### Issue closed
None

#### Tier 1:
 - [x] eda_OIE120km_WarmStart

#### Tier 3 (optional):
 - [x] ForecastFromGFSAnalysesVariableMesh


scenarios/defaults/model.yaml
```diff
@@ -7,15 +7,79 @@ model:
...
+    #  RadiationCloud: str
+    #  RadiationLW: str
+    #  RadiationSW: str
+    #  SfcLayer: str
+    #  LSM: str
+    60-3km:
+      nCells: 835586
+      TimeStep: 20.0
+      DiffusionLengthScale: 3000.0
+      RadiationLWInterval: 00:30:00
```
RGNystrom: Based on other's setting for regional MPAS with the 3 km horizontal grid spacing, I recommend changing the default LW and SW Interval to `config_radtlw_interval     = '00:03:00'`.

scenarios/defaults/model.yaml
```diff
@@ -7,15 +7,79 @@ model:
...
+      RadiationSWInterval: 00:30:00
+      PhysicsSuite: mesoscale_reference
```
RGNystrom: I think changing PhysicsSuite to 'convection_permitting' would be best.

scenarios/defaults/model.yaml
ibanos90: Thanks for the suggestion @RGNystrom! Do you have an idea about the walltime after calling the radiation scheme more frequently?

scenarios/defaults/model.yaml
ibanos90: That's a good point. I can definitely change it. I just set this as the default given the configuration we are using in our DI experiments. Note that these variables can be changed at any time in the scenario yaml file.

mos3r3n:
If we want to run variable mesh experiments, we need to add the section 'staticstream' and define 'filePrefix' under it. I am still trying it and let you know the update.

bin/PrepJEDI.csh
```diff
@@ -182,6 +186,20 @@ foreach NamelistFile_ ($NamelistFileList)
...
+sed -i 's@micropScheme@'${Microphysics}'@' $NamelistFile_
+sed -i 's@convectionScheme@'${Convection}'@' $NamelistFile_
+sed -i 's@pblScheme@'${PBL}'@' $NamelistFile_
+sed -i 's@gwdoScheme@'${Gwdo}'@' $NamelistFile_
+sed -i 's@radtCldScheme@'${RadiationCloud}'@' $NamelistFile_
+sed -i 's@radtLWScheme@'${RadiationLW}'@' $NamelistFile_
+sed -i 's@radtSWScheme@'${RadiationSW}'@' $NamelistFile_
+sed -i 's@sfcLayerScheme@'${SfcLayer}'@' $NamelistFile_
+sed -i 's@lsmScheme@'${LSM}'@' $NamelistFile_
+
```
mos3r3n: These modifications should be done for the namelists of all meshes. Move it before the "end" in Line 183.

ibanos90:
> If we want to run variable mesh experiments, we need to add the section 'staticstream' and define 'filePrefix' under it. I am still trying it and let you know the update.

You mean for DA experiments with variable mesh?

mos3r3n:
> > If we want to run variable mesh experiments, we need to add the section 'staticstream' and define 'filePrefix' under it. I am still trying it and let you know the update.
> 
> You mean for DA experiments with variable mesh?

Yes! And we should be careful with its prefix "x1". For 60-3 km mesh, its prefix is x20.

bin/PrepJEDI.csh
ibanos90: That's right, I copied this from my actual local branch, but missed that. Thanks for the catch!

ibanos90:
> > > If we want to run variable mesh experiments, we need to add the section 'staticstream' and define 'filePrefix' under it. I am still trying it and let you know the update.
> > 
> > 
> > You mean for DA experiments with variable mesh?
> 
> Yes! And we should be careful with its prefix "x1". For 60-3 km mesh, its prefix is x20.

I get that, but the goal of this PR is to make the variable mesh work for the forecast. Do you mind opening a PR after this to include changes for DA?

Regarding the prefix, for simplicity I changed the name of the mesh, graph, and static files to be `x1`, then I did not need to add another variable for that. Nonetheless, I agree the most correct thing is to keep it as `x20`. Will add that.

mos3r3n:
> > > > If we want to run variable mesh experiments, we need to add the section 'staticstream' and define 'filePrefix' under it. I am still trying it and let you know the update.
> > > 
> > > 
> > > You mean for DA experiments with variable mesh?
> > 
> > 
> > Yes! And we should be careful with its prefix "x1". For 60-3 km mesh, its prefix is x20.
> 
> I get that, but the goal of this PR is to make the variable mesh work for the forecast. Do you mind opening a PR after this to include changes for DA?
> 
> Regarding the prefix, for simplicity I changed the name of the mesh, graph, and static files to be `x1`, then I did not need to add another variable for that. Nonetheless, I agree the most correct thing is to keep it as `x20`. Will add that.

I see. If this PR doesn't handle DA, I will submit a new one for DA cycling. Thanks!

mos3r3n:
UngribExt*    --    1   1  235gb 00:05 Q   --  
ObsToIODA*    --    1   1  125gb 00:15 Q   --

I suggest using 128 cores per node for those jobs on Derecho to reduce the queue time.

ibanos90:
> UngribExt* -- 1 1 235gb 00:05 Q -- ObsToIODA* -- 1 1 125gb 00:15 Q --
> 
> I suggest using 128 cores per node for those jobs on Derecho to reduce the queue time.

I made those changes in latest commits. Thanks for testing it @mos3r3n!

scenarios/defaults/model.yaml
ibanos90: Changed it as suggested

scenarios/defaults/model.yaml
ibanos90: Changed it as suggested. The walltime does not change much by forecast hour, i.e., from 4min to ~5min. I made the corresponding change in the job requirements

mos3r3n (approved):

PR #306, opened 2024-05-14
Add support for log levels.

jim-p-w:
### Description
Different log levels will result in more or fewer log messages.
The level is controlled by the CYLC_DEBUG environment variable.
A value of 0 has the fewest messages, a value of 3 is the most verbose.
Default value (if CYLC_DEBUG is not set) is level 1.

To set the logging verbosity, before invoking Run.py or test.csh set CYLC_DEBUG
bash: `export CYLC_DEBUG=0`
tcsh: `setenv CYLC_DEBUG 0`

level 1 will show 
```
#########################################################################
Running the scenario: test/testinput/3dvar_OIE120km_WarmStart.yaml
#########################################################################
Experiment: ======================================================================
Experiment: Setting up a new suite
Experiment:   SuiteName: jwittig_3dvar_OIE120km_WarmStart_TEST
Experiment:   mainScriptDir: /glade/derecho/scratch/jwittig/pandac/jwittig_3dvar_OIE120km_WarmStart_TEST/MPAS-Workflow
Experiment:   ExperimentDirectory: /glade/derecho/scratch/jwittig/pandac/jwittig_3dvar_OIE120km_WarmStart_TEST
Experiment: ======================================================================
./submit.csh cylc version: 8.2.2
./submit.csh (INFO): checking if a suite with the same name is already running
./submit.csh cylc install --run-name=jwittig_3dvar_OIE120km_WarmStart_TEST
Valid for cylc-8.2.2

 ▪ ■  Cylc Workflow Engine 8.2.2
 ██   Copyright (C) 2008-2023 NIWA
▝▘    & British Crown (Met Office) & Contributors
```

level 0 will show the same, excluding the `Experiment` lines
level 2 will show similar output you see before this PR.
level 3 will show a little more than level 2.

### Issue closed

Closes #(if applicable)

#### Tier 1:
 - [x] eda_OIE120km_WarmStart
    - This failed due to missing data files for the ExtendedMeanFC steps. File `x1.655362.graph.info.part.128` is missing.

#### Tier 2 (optional):
 - [ ] GenerateGFSAnalyses
 - [ ] GenerateObs

### DIFF
--- a/Run.py
+++ b/Run.py
@@ -44,9 +44,9 @@ def main():
   run.execute()
 
 
-class Run():
+class Run(Logger):
   def __init__(self):
-    self.logPrefix = self.__class__.__name__+': '
+    super().__init__()
 
     # Parse command line
     ap = argparse.ArgumentParser()
@@ -77,7 +77,7 @@ class Run():
       print("Running the scenario: "+scenarioFile)
       print("#########################################################################")
 
-      self.clean()
+      self.clean(self)
 
       scenario = Scenario(scenarioFile)
       scenario.initialize()
@@ -88,12 +88,11 @@ class Run():
       suite = SuiteLookup(suiteName, scenario.getConfig())
       suite.submit()
 
-    self.clean()
+    self.clean(self)
 
   @staticmethod
-  def clean():
-    logger = Logger()
-    logger.log('cleaning up auto-generated files...', level=logger.MSG_DEBUG)
+  def clean(self):
+    self.log('cleaning up auto-generated files...', level=self.MSG_DEBUG)
 
     for g in [
       "config/auto/*.csh",
--- a/Run.py
+++ b/Run.py
@@ -26,6 +26,7 @@ import subprocess
 
 # local modules
 from initialize.config.Config import Config
+from initialize.config.Logger import Logger
 from initialize.config.Scenario import Scenario
 from initialize.suites.SuiteBase import SuiteLookup
 
@@ -91,7 +92,8 @@ class Run():
 
   @staticmethod
   def clean():
-    print('cleaning up auto-generated files...')
+    logger = Logger()
+    logger.log('cleaning up auto-generated files...', level=logger.MSG_DEBUG)
 
     for g in [
       "config/auto/*.csh",
--- a/initialize/config/Component.py
+++ b/initialize/config/Component.py
@@ -9,17 +9,22 @@
 
 from collections.abc import Iterable
 
+from initialize.config.Logger import Logger
 from initialize.config.Config import Config
 from initialize.config.TaskFamily import CylcTaskFamily
 from initialize.config.SubConfig import SubConfig
 
-class Component():
+class Component(Logger):
+
   defaults = None
   workDir = None
   requiredVariables = {}
   optionalVariables = {}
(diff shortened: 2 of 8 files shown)


amstokely:
Looks good to me! I would add type hints to the methods you modified/added, as they make python a lot easier to follow (at least  imo).

amstokely (approved):

amstokely (approved):
Looks good to me! I would add type hints to the methods you modified/added, as they make python a lot easier to follow (at least imo).

ibanos90 (changes requested):
This is a very helpful PR! As we talked, it would be good to add to the description how to use/change this variable for the logging level. From the code is wasn't clear how I should specify it. Thanks!

jim-p-w:
I've added some information to the PR description which explains how to set the logging level.

ibanos90:
> I've added some information to the PR description which explains how to set the logging level.

@jim-p-w Thanks for adding the info to the PR description! I tested all the options and it's very helpful to have this option to add or reduce debugging level when submitting the cylc suites. One question though, do we know if this debugging level will also give us more/less info in the cylc-run directory outputs from cylc? Just curious about it.

Another thing, I noted that in the reference tests you mentioned about the graph file missing. This is because is using `/glade/work/liuz/pandac/prepare_mpas/mpas_static_30km` as the path for the `GraphInfoDir` variable, but this folder does not exist anymore. This is also used in `scenarios/VerifyingGFSAnalysesFromRDA.yaml`. I would suggest to remove the path from the scenario yaml files and use the default path for `GraphInfoDir` specified in `data/Model.py`.

We are using graphs from `/glade/work/duda/static_moved_to_campaign`  (in some scenario yamls) and `/glade/campaign/mmm/parc/liuz/pandac_common/static_from_duda` (default path), however, my guess is that we could use just `/glade/campaign/mmm/parc/liuz/pandac_common/static_from_duda`. Can you confirm @liujake? I cannot remember if there was a particular reason why we would keep using `/glade/work/duda/static_moved_to_campaign`

jim-p-w:
@ibanos90 This code change only impacts what is written to the console when starting a scenario.

jim-p-w:
@ibanos90 Regarding the `GraphInfoDir` changes, I would prefer to make those changes under a separate PR, as they aren't related to these logging changes. (I am happy to create that PR after we decide what changes are needed for that.) Should I open an issue for the `GraphInfoDir` changes?

ibanos90:
> @ibanos90 This code change only impacts what is written to the console when starting a scenario.

That was my guess but good to know for sure. Thanks!

ibanos90:
> @ibanos90 Regarding the `GraphInfoDir` changes, I would prefer to make those changes under a separate PR, as they aren't related to these logging changes. (I am happy to create that PR after we decide what changes are needed for that.) Should I open an issue for the `GraphInfoDir` changes?

That would be great! I also noted that we use two different folders for the external analysis. After some info exchange with @liujake, we may want to just use the external analysis under directories such as `/glade/campaign/mmm/parc/liuz/pandac_common/30km/30km_GFSANA_single`. Maybe this could be included in the PR for the GraphInfo path as part of some cleanup/organization.

ibanos90 (approved):

PR #311, opened 2024-06-04
Use 48h late prepbufr file as default

junmeiban:
### Description
@liujake noticed the obs number of aircraft is very low when we use observation from `prepnr` directory.
This PR uses the prepbufr file with 48h time lag as default (in RDA `prep48h` directory). If the file does not exist in `prep48` directory, we will use prepbufr file from `prepnr` directory.

### Files modified:
M   bin/GetObs.csh
### Tests completed
 - [x] GenerateObs


ibanos90 (approved):
Thanks @junmeiban for adding this PR! I ran the test and it works successfully.

PR #313, opened 2024-07-10
[Draft] Updates for using the public MPAS-Dev/MPAS_Model in mpas-bundle.

mos3r3n:
### Description
This update is for implementing the public MPAS model in mpas-bundle. 

- Change static to invariant in two-stream files.
- Modify noah to sf_noah in model forecasting namelist.
- For DA, instead of copying the background file to the analysis file for overwriting, a new analysis file will be generated by the mpas-jedi. After DA, the analysis variables in the analysis file are overwritten to another existing mpasout file, copied from the background file. This treatment is due to the bug that the analysis variables cannot be directly written into an existing file in MPAS-JEDI. Now, only variational DA works. EnKF run mail fail. (This bug also appears in the variable-resolution mesh in JCSDA-internal MPAS).

#### Tier 1:
 - [ ] eda_OIE120km_WarmStart

#### Tier 2 (optional):
 - [ ] GenerateGFSAnalyses
 - [ ] GenerateObs

### DIFF
--- a/bin/PrepJEDI.csh
+++ b/bin/PrepJEDI.csh
@@ -1162,8 +1162,8 @@ if ("$ArgAppType" == variational) then
     set an = $CyclingDAOutDirs[$member]
     mkdir -p ${an}
     set anFile = ${an}/${ANFilePrefix}.$thisMPASFileDate.nc
-    rm ${anFile}
-    cp -v ${bgFile} ${anFile}
+    rm ${anFile}*
+    cp -v ${bgFile} ${anFile}.bak
 
     @ member++
   end
@@ -1223,8 +1223,8 @@ else if ("$ArgAppType" == enkf) then
     set an = $CyclingDAOutDirs[$member]
     mkdir -p ${an}
     set anFile = ${an}/${ANFilePrefix}.$thisMPASFileDate.nc
-    rm ${anFile}
-    cp -v ${bgFile} ${anFile}
+    rm ${anFile}*
+    cp -v ${bgFile} ${anFile}.bak
 
     @ member++
   end
--- a/bin/Variational.csh
+++ b/bin/Variational.csh
@@ -115,6 +115,14 @@ if ( $status != 0 ) then
   exit 1
 endif
 
+# Update state variables of analysis
+set an = $CyclingDAOutDirs[$ArgMember]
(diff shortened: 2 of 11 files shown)


jim-p-w:
I verified running against MPAS-Dev/MPAS-Model without the PIO workaround in `bin/Variational.csh`, `config/tools.csh`, and `tools/update_analysis_states.py` works.  Can you remove those changes from this PR?

mos3r3n:
Hi Jim,

Thanks for your efforts on testing it.
Have you ever tried it with variable-resolution mesh? We had problem when over-writing the analysis variables to an existing analysis file. If this issue still exists, I will keep these modifications but specifically for the variable-resolution mesh.

Thanks!

Tao


From: Jim W ***@***.***>
Date: Monday, July 15, 2024 at 1:20 PM
To: NCAR/MPAS-Workflow ***@***.***>
Cc: Tao Sun ***@***.***>, Author ***@***.***>
Subject: Re: [NCAR/MPAS-Workflow] [Draft] Updates for using the public MPAS-Dev/MPAS_Model in mpas-bundle. (PR #313)

I verified running against MPAS-Dev/MPAS-Model without the PIO workaround in bin/Variational.csh, config/tools.csh, and tools/update_analysis_states.py works. Can you remove those changes from this PR?

liujake:
@jim-p-w @mos3r3n I suggest we keep this 'workaround' feature branch for future reference in case we need something similar in a future time.

@jim-p-w You may create a separate branch and PR by removing the workaround changes and then make a PR.

jim-p-w:
It turns out the workaround is needed for variable mesh experiments. 

The workaround is needed when using either the jcsda internal mpas-model or the public mpas-model code. Given that, it seems the problem with variable mesh experiments is not due to the PIO bug in the public mpas-model code.

mos3r3n:
> It turns out the workaround is needed for variable mesh experiments.
> 
> The workaround is needed when using either the jcsda internal mpas-model or the public mpas-model code. Given that, it seems the problem with variable mesh experiments is not due to the PIO bug in the public maps-model code.

It is very hard to check the issue if not setting OOPS_TRACE = 1. You can see that the status of an.*.nc keeps updating but never finishes. We need take some time to fix this bug.

PR #314, opened 2024-07-17
[Draft] Support using the public mpas-model release 8.2

jim-p-w:
This PR contains changes to switch from using the internal JCSDA mpas-model to the public mpas-model release 8.2.
It includes
 - Change stream names from "static" to "invariant".
 - Change lsm scheme name from "noah" to "sf_noah"

It is based on changes in https://github.com/NCAR/MPAS-Workflow/pull/313, authored by @mos3r3n.

#### Tier 1:
 - [ ] eda_OIE120km_WarmStart

#### Tier 2 (optional):
 - [ ] GenerateGFSAnalyses
 - [ ] GenerateObs

### DIFF
--- a/config/mpas/forecast/streams.atmosphere
+++ b/config/mpas/forecast/streams.atmosphere
@@ -1,5 +1,5 @@
 <streams>
-<immutable_stream name="static"
+<immutable_stream name="invariant"
                   type="input"
                   precision="{{PRECISION}}"
                   filename_template="{{StaticFieldsPrefix}}.{{nCells}}.nc"
--- a/config/mpas/hofx/streams.atmosphere
+++ b/config/mpas/hofx/streams.atmosphere
@@ -1,5 +1,5 @@
 <streams>
-<immutable_stream name="static"
+<immutable_stream name="invariant"
                   type="input"
                   precision="{{PRECISION}}"
                   filename_template="{{StaticFieldsPrefix}}.{{nCells}}.nc"
--- a/config/mpas/rtpp/streams.atmosphere
+++ b/config/mpas/rtpp/streams.atmosphere
@@ -1,5 +1,5 @@
 <streams>
-<immutable_stream name="static"
+<immutable_stream name="invariant"
                   type="input"
                   precision="{{PRECISION}}"
                   filename_template="{{StaticFieldsPrefix}}.{{nCells}}.nc"
--- a/config/mpas/saca/streams.atmosphere
+++ b/config/mpas/saca/streams.atmosphere
@@ -1,5 +1,5 @@
 <streams>
-<immutable_stream name="static"
+<immutable_stream name="invariant"
                   type="input"
                   filename_template="{{StaticFieldsPrefix}}.{{nCells}}.nc"
                   precision="{{PRECISION}}"
--- a/config/mpas/variational/streams.atmosphere
+++ b/config/mpas/variational/streams.atmosphere
@@ -1,5 +1,5 @@
 <streams>
-<immutable_stream name="static"
+<immutable_stream name="invariant"
                   type="input"
                   precision="{{PRECISION}}"
                   filename_template="{{StaticFieldsPrefix}}.{{nCells}}.nc{{TemplateFieldsMember}}"
--- a/initialize/framework/Build.py
+++ b/initialize/framework/Build.py
@@ -40,11 +40,9 @@ class Build(Component):
     # set system dependent defaults before invoking Component ctor
     system = os.getenv('NCAR_HOST')
     if system == 'derecho':
-      if config._bundle_dir != None:
-        self.variablesWithDefaults['mpas bundle'] = [config._bundle_dir, str]
-      else:
-        self.variablesWithDefaults['mpas bundle'] = \
-          ['/glade/campaign/mmm/parc/ivette/pandac/codeBuild/mpasBundle_saca_dev_10Jun2024/build_SP', str]
+      self.variablesWithDefaults['mpas bundle'] = \
+        ['/glade/work/jwittig/repos1/mpas-bundle-model-8.1/build-gnu-1p', str]
+        #['/glade/u/home/taosun/work/Derecho/JEDI/mpas-bundle-dev/build', str]
       self.variablesWithDefaults['bundle compiler used'] = ['gnu-cray', str,
         ['gnu-cray', 'intel-cray']]
       self.variablesWithDefaults['forecast directory'] = ['bundle', str]
--- a/scenarios/defaults/model.yaml
+++ b/scenarios/defaults/model.yaml
@@ -36,7 +36,7 @@ model:
       RadiationLW: rrtmg_lw
       RadiationSW: rrtmg_sw
       SfcLayer: sf_mynn
-      LSM: noah
+      LSM: sf_noah
     30km:
       meshRatio: 1
       nCells: 655362
@@ -53,7 +53,7 @@ model:
       RadiationLW: rrtmg_lw
       RadiationSW: rrtmg_sw
       SfcLayer: sf_monin_obukhov
-      LSM: noah
+      LSM: sf_noah
     60km:
       meshRatio: 1
       nCells: 163842
@@ -70,7 +70,7 @@ model:
       RadiationLW: rrtmg_lw
       RadiationSW: rrtmg_sw
       SfcLayer: sf_monin_obukhov
-      LSM: noah
+      LSM: sf_noah
     120km:
       meshRatio: 1
       nCells: 40962
@@ -87,4 +87,4 @@ model:
       RadiationLW: rrtmg_lw
       RadiationSW: rrtmg_sw
       SfcLayer: sf_monin_obukhov
-      LSM: noah
+      LSM: sf_noah


ibanos90:
Hi @jim-p-w, @liujake, @amstokely, @mos3r3n, and @byoung-joo, I ran the reference tests with this feature branch and got some interesting results:

1. Dual resolution 30km-60km tests failed with same issue as we have seen for the variable mesh, i.e., gets stuck after minimization and QCs after analysis but cannot overwrite the background files with updated analysis variables

2. Tests for ColdStart initialization and ForecastFromGFSAnalysesMPT failed in the creation of the external analysis with the following error message in files such as `log.init_atmosphere.0013.err`:
```
ERROR: At least one fields to be read from the 'input' stream is dimensioned
ERROR: by 'nSoilComps', but the 'nSoilComps' dimension is not defined
ERROR: in the file x1.40962.static.nc
CRITICAL ERROR: Please check the input file(s) to be read by the 'input' input stream.
```
and in the `log.init_atmosphere.0000.out`
```
 Reading dimensions from input streams ...

 ----- reading dimensions from stream 'input' using file x1.40962.static.nc
        nCells = 40962
        nEdges = 122880
        nVertices = 81920
        TWO = 2
        maxEdges = 10
        maxEdges2 = 20
        vertexDegree = 3
        R3 = 3
        nMonths = 12
        nSoilComps *** not found in stream ***

ERROR: At least one fields to be read from the 'input' stream is dimensioned
```
Do we need to update the static file to one created with this new version of the model, or make any additional change?

Tests with single resolution for DA, such as 120km and with WarmStart for 3denvar, getkf and eda, passed successfully. In case you want to take a look, please find the tests here: /glade/derecho/scratch/ivette/pandac-PR314

byoung-joo:
I wonder if this is related to https://github.com/MPAS-Dev/MPAS-Model/pull/1194.  From that PR, `nSoilComps` is only related to `sf_noahmp`, while in our workflow we use `sf_noah`.

byoung-joo:
Ah, https://github.com/MPAS-Dev/MPAS-Model/pull/1194 is related to `core_atmosphere`.

I realized that the error message Ivette sees is related to `core_init_atmosphere`.
Yes, I think we need to re-generate our existing MPAS static files, such as it can be also used for `sf_noahmp` option.
Can you confirm this, @mgduda ?

byoung-joo:
The error message is a bit confusing because from our existing static files, there is no variable with a dimension `nSoilComps`...

ibanos90:
> The error message is a bit confusing because from our existing static files, there is no variable with a dimension `nSoilComps`...

Exactly! That's why I wondered if we needed to update the static file.

liujake:
@mos3r3n when you run 60km-3km test with public model 8.2.x code, did you generate 'invariant' file, or still using 'static' file (linked to a init.nc file)?

mos3r3n:
> @mos3r3n when you run 60km-3km test with public model 8.2.x code, did you generate 'invariant' file, or still using 'static' file (linked to a init.nc file)?

I have not tested v8.2.x yet. In my v8.1.x run, the invariant file is linked from the init.nc file, which is generated by myself.

liujake:
In that case, we need a cleaner test by using a real 'invariant' file produced using 8.2.x code. There is a chapter 8.3 (if I remember correctly) for generating invariant file, from the latest model user guide. @jim-p-w I guess you also did not generate an 'invariant' file for your 120km test. So I consider our 'public' cycling tests so far are not rigorous. We need to re-do the 'public' tests with an 'invariant' file for all scenarios.

liujake:
https://www2.mmm.ucar.edu/projects/mpas/mpas_atmosphere_users_guide_8.2.0.pdf

liujake:
@jim-p-w It seems to me that we should need to modify somewhere in the workflow script to link a real 'invariant' file, instead of the old 'static.nc' linked to an 'init.nc' file. But I only see 'static' stream is changed to 'invariant' stream in configuration files, but looks like still using 'static.nc' file for the 'invariant' stream. I think we'd better make it logically consistent to use an 'invariant.nc' file for the 'invariant' stream. 'static.nc' and 'invariant.nc' are different things in 8.1/8.2 notation. We need to be rigorous and remove the use of 'static.nc' file in the workflow.

jim-p-w:
@liujake I am looking into the static/invariant issue. Michael's initial thought was using the `static` file as the `invariant` file should be OK. Apparently the static stream was never officially supported and there's no documentation for it.

liujake:
We can assume our use of 'static' file produces the same results as using 'invariant' file. But we need to confirm it by using 'invariant' file. Take your time to learn first to generate necessary 'invariant' file(s), which is something new to us. I did not make this very clear from the beginning of testing 'public' mpas-bundle code.

ibanos90:
Hi all, an update on my recent tests using Jim's build code with MPAS-Model v8.2.1:

**30km-60km:**
 - Created static, invariant, init and 1st background (following the user's guide). Ran test and DA worked successfully!!!
 Experiment is almost finished!
Please check: /glade/derecho/scratch/ivette/pandac/ivette_3dhybrid_O30kmIE60km_public_MPAS-modelv8.2.1/CyclingDA/2018041500/run/mem001

 Two things here for the success of this test:
1.  For our DA, we use the init file as our static/invariant file, therefore, here we have to regenerate init files for this experiment with new public model version to work.

 2. I tried using new init file with old 1st background file and did NOT work, therefore, the 1st background files also need to be regenerated.

**60-3km-60km:**
 - Created static, invariant, init and 1st background. Ran test and DA did NOT work. Differently than before, it now shows an error message in `jedi.log.all/jedi.log`: 
```
 --> write_fields: MPAS_stream_mgr_write failed ierr=          -3

MPICH ERROR [Rank 3] [job id 31e52196-bc35-4d1c-93ae-eec8a32ab078] [Sat Jul 20 23:34:11 2024] [dec1985] - Abort(1) (rank 3 in comm 0): application called MPI_Abort(MPI_COMM_WORLD, 1) - process 3
```
It also shows this message in `log.atmosphere.0255.d0002.err`
```
ERROR: MPAS IO Error: Undefined field
```
Please check: /glade/derecho/scratch/ivette/pandac/ivette_3dhybrid_O60-3kmIE60km_TEST_Modelv8.2.1/CyclingDA/2018041500/run/mem001

**Important**
- While using Jim's build code, I noted that the new INIT files created do not contain diagnostics for 3D pressure or winds at cell centers (uReconstructZonal and uReconstructMeridional), which are used for our model space verification. Please check these lines: https://github.com/amstokely/MPAS-Model/blob/c8f4fe737b80990db3143eecf01b264f3db1ebe8/src/core_atmosphere/Registry.xml#L526-L572
I tried to make a new compilation adding these variables in that section for 'input' but didn't work, I may have missed something. May need some help adding these. In the meantime, I will try using old init files for model space verification.

ibanos90:
I forgot to mention, I created a new branch changing all references to old static files and renaming then as invariant, which I have been using for my experiments: https://github.com/NCAR/MPAS-Workflow/tree/feature/changes4publicMPAS-8.2.1

liujake:
@ibanos90 I want to make sure I understand you correctly: during the cycling of 30km-60km run, you used the new 'invariant.nc' files (one for 30km, one for 60km) in place of the old 'static.nc' files (which is linked to an 'init.nc' file)? i.e., during the cycling, you have mpasout file along with invariant file.

ibanos90:
> @ibanos90 I want to make sure I understand you correctly: during the cycling of 30km-60km run, you used the new 'invariant.nc' files (one for 30km, one for 60km) in place of the old 'static.nc' files (which is linked to an 'init.nc' file)? i.e., during the cycling, you have mpasout file along with invariant file.

That's correct, but I created new init file only for 30km in this case and worked

ibanos90:
Hi all, my last test using same physical parameterizations as the ensemble members and 1st background (i.e., mesoscale reference suite) succeeded for the DA dual resolution @60-3km-60km! 

If I use a different configuration in the parameterizations the test hangs until the walltime is exceeded. This wasn't noted before because for 30km-60km we use mesoscale reference for everything (ensemble forecast members, 1st background, analysis, subsequent forecasts).

liujake:
@ibanos90 Good catch! So that was an issue of physics inconsistency between ensemble input and high-res. background/analysis.

ibanos90:
> @ibanos90 Good catch! So that was an issue of physics inconsistency between ensemble input and high-res. background/analysis.

Exactly!

liujake:
Or inconsistency between 2 template files at different resolution could also played a role in the overwriting issue: one at 60km-3km (convection-permitting suite), one at 60km (mesoscale suite).

mos3r3n:
> > @ibanos90 Good catch! So that was an issue of physics inconsistency between ensemble input and high-res. background/analysis.
> 
> Exactly!

Hi @ibanos90. I noticed that in DA and forecasting, the init.nc was still linked as invariant.nc. Have you ever tried just using the invariant.nc? I think we need to do that eventually.

ibanos90:
> > > @ibanos90 Good catch! So that was an issue of physics inconsistency between ensemble input and high-res. background/analysis.
> > 
> > 
> > Exactly!
> 
> Hi @ibanos90. I noticed that in DA and forecasting, the init.nc was still linked as invariant.nc. Have you ever tried just using the invariant.nc? I think we need to do that eventually.

Hi @mos3r3n, I am not sure that's what we want to do. We link the init file as we do the dual resolution approach in the cycling DA. In the past, I have tried linking the static file and it doesn't work for the DA. @byoung-joo can you provide more guidance on this?

liujake:
There might be misunderstanding here. The right way with 8.2.1, as I mentioned in earlier conversation, is to link the 'invariant.nc' file to 'invariant' stream file. The 'static.nc' file or 'init.nc' file should no longer be involved in the cycling. We should only need mpasout file and invariant file in the cycling. That is the main purpose we switch to the new 8.2.1 code. The old way of pretending init.nc file as static.nc file is not a clean way to do and is confusing.

liujake:
If you link 'init.nc' to 'invariant.nc' in the cycling workflow, it is like you pretend init.nc as invariant.nc. But we should use actual invariant.nc file you generated, not init.nc file.

ibanos90:
That's a good point. I tried the old way. I didn't realize this change, sorry about that.

PR #318, opened 2024-07-24
Bug fix for surface obs generation.

junmeiban:
### Description
This PR is a bug fix for generating surface obs:

- For surface obs, our purpose is to use line 112 to generate surface obs with "-noqc" option to keep as more observation as possible,  and use this file (generated by line 112) to replace previous one which generated by line 107 without "-noqc" option.  

- While the command at line 112 is not executed because the log file is pointed to the wrong path.

### Issue closed

Closes #(if applicable)

### Tests completed
 - [x] GenerateObs

### DIFF
--- a/bin/ObsToIODA.csh
+++ b/bin/ObsToIODA.csh
@@ -109,7 +109,7 @@ foreach gdasfile ( *"gdas."* )
      mkdir -p sfc
      cd sfc
      ln -sfv ${obs2iodaBuildDir}/${obs2iodaEXE} ./
-     ./${obs2iodaEXE} ${noGSIQCFilters} ../${gdasfile} >&! logs/log-converter_sfc
+     ./${obs2iodaEXE} ${noGSIQCFilters} ../${gdasfile} >&! ../logs/log-converter_sfc
      # replace surface obs file with file created without additional QC
      mv -f sfc_obs_${thisCycleDate}.h5 ../sfc_obs_${thisCycleDate}.h5
      cd ..


liujake:
@junmeiban  How many more Ps obs from '-noqc' option?

junmeiban:
> @junmeiban How many more Ps obs from '-noqc' option?

For 2023 experiments, I checked several dates and there is no more obs with '-noqc' option (see the following figures). For 2018 experiments, there are over 4K more Ps obs with '-noqc' option. 
For our 2018 experiments, all observations are generated by shell scripts not by workflow, so our cycling experiment (2018) are using correct Ps obs. 

20230704 without bugfix:
distri_stationPressure_sfc_obs_2023070400_all_wofix

20230704 with bugfix:
distri_stationPressure_sfc_obs_2023070400_all_fix

20180415 without bugfix:
distri_stationPressure_sfc_obs_2018041500_all_wofix

20180415 with bugfix:
distri_stationPressure_sfc_obs_2018041500_all_fix

junmeiban:
From the above figures, there are less (more) Ps obs over ocean (Brazil) for 2023, comparing to 2018.

ibanos90 (approved):
Good catch! Thanks @junmeiban!

liujake:
Do you know what exactly QC is applied if we turn on QC for Ps processing?

junmeiban:
For default ( ‘-qc’ =true), it’s corresponding to `apply_gsi_qc` in the source code and subroutine `filter_obs_conv` will be called. There are criterion in `filter_obs_conv` for sfc obs, such as pressure, observation quality marker(qm) to filter out obs:
https://github.com/NCAR/obs2ioda/blob/36f6e5100c55673ff3f3159c65532bba945c8a05/obs2ioda-v1/src/prepbufr_mod.f90#L915

liujake:
Can you remind me why we want to turn QC off, specifically for Ps in the conversion step, but turn QC on for other obs types in prepbufr?

liujake (approved):
Looks good.

junmeiban:
It's our experience from 2018 experiments. In this way we got more Ps + use background check when run DA. And we indeed got improvements. 

> Can you remind me why we want to turn QC off, specifically for Ps in the conversion step, but turn QC on for other obs types in prepbufr?

zhumingying (approved):

PR #323, opened 2024-09-16
Update CMakeLists for mpas-bundle v3.0.0

ibanos90:
### Description
Update CMakeLists for mpas-bundle v3.0.0 (taken from [CMakeLists.txt](https://github.com/JCSDA-internal/mpas-bundle/blob/release/3.0.0/CMakeLists.txt)) so users can refer to these hash numbers in the future.

### DIFF
--- a/build/CMakeLists.txt
+++ b/build/CMakeLists.txt
@@ -1,4 +1,4 @@
-# (C) Copyright 2017-2020 UCAR
+# (C) Copyright 2017-2024 UCAR
 #
 # This software is licensed under the terms of the Apache Licence Version 2.0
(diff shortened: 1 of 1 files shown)


build/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-# (C) Copyright 2017-2023 UCAR
+# (C) Copyright 2017-2020 UCAR
```
liujake: We may want to change this to 2024

ibanos90: Good point. Done!

liujake (approved):

PR #337, opened 2024-12-17
Fix concat task dependencies for Variational and EnKF

ibanos90:
### Description
This PR fixes the dependency for the `concat` task in the `Variational` and `EnKF` applications. The `HofX` was fixed some time ago but I missed fixing these two as well.

### Issue closed
None

#### Tier 1:
 - [x] eda_OIE120km_WarmStart


junmeiban (approved):
Tested by using 3denvar scenario for several cycles. The concatenated files were successfully generated, and the workflow moved to the next cycle without issues. Many thanks for fixing this.

ibanos90:
> Tested by using 3denvar scenario for several cycles. The concatenated files were successfully generated, and the workflow moved to the next cycle without issues. Many thanks for fixing this.

Thanks so much for testing, @junmeiban!

Issue #215, opened 2023-04-05
Make VerifyObs work with EnKF output


jjguerrette:
VerifyObs only works with HofX and Variational feedback files.  I will not be able to make it work with EnKF feedback files (very different HDF5 group naming conventions than for HofX and Variational) before leaving.  I'm adding this issue so others are aware.

PR #338, opened 2025-01-07
Enhancement of GETKF/LETKF ensemble DA capability.

mos3r3n:
### Description
This PR introduces various new features in LETKF/GETKF:
1. use the linear operator as default for more efficient observer step;
3. Enable inflation within LETKF/GETKF itself, instead of using the RTPP application;
4. Use 'reduce size space' for the satellite data thinning section to reduce memory and IO burden. It also benefits the variational DA.
5. Enable offline bias correction in LETKF/GETKF.
6. Calculate HofX of ensemble analyses (EnKFDiagOMA), separately from the Observer and Solver steps.

Note: The two-step strategy is still used, due to the poor efficiency of using 'Halo' distribution for ensemble HofX calculation; EnKFDiagOMA is now still a part of ensemble DA, making it a dependency for ensemble forecasts (To be improved).

### Issue closed

Closes #(if applicable)

#### Tier 1:
 - [x] eda_OIE120km_WarmStart

#### Tier 2 (optional):
 - [ ] GenerateGFSAnalyses
 - [ ] GenerateObs


bin/EnKF.csh
```diff
@@ -75,6 +75,8 @@ grep 'Run: Finishing oops.* with status = 0' solver.log
 if ( $status != 0 ) then
   echo "ERROR in $0 : enkf solver failed" > ./FAIL
   exit 1
+else
+  rm solver.log.0*
```
ibanos90: We usually keep each processor log for debugging. Should we remove them?

bin/EnKFObserver.csh
```diff
@@ -44,46 +45,68 @@ set myYAML = ${self_WorkDir}/${appyaml}
...
 # Check status
 # ============
-grep 'Run: Finishing oops.* with status = 0' observer.log
+grep 'Run: Finishing oops.* with status = 0' ${appName}.log
 if ( $status != 0 ) then
-  echo "ERROR in $0 : enkf observer failed" > ./FAIL
+  echo "ERROR in $0 : enkf ${appName} failed" > ./FAIL
   exit 1
+else
+  rm ${appName}.log.0*
```
ibanos90: We usually keep each processor log for debugging. Should we remove them?

bin/PrepJEDI.csh
```diff
@@ -917,14 +941,19 @@ else if ("$ArgAppType" == enkf) then
   cp $appyaml prevPrep.yaml
   set prevYAML = $appyaml
 
-  # Analysis directory
+  # Analysis diuseLinearOperatorrectory
```
ibanos90: Is this a typo?

bin/EnKF.csh
mos3r3n: > We usually keep each processor log for debugging. Should we remove them?

In my opinion, those *.log.all files already contain error info. That's why I prefer to delete those *.log.0* files. If you think those files are useful, I will keep them in our working directories.

bin/PrepJEDI.csh
mos3r3n: > Is this a typo?

Yes, it is. I will correct it.

bin/EnKF.csh
liujake: I agree removing log files for each processor will make working dir much cleaner. I usually delete those files manually when archiving folders to campaign storage and just keep jedi.log for variational runs. Those processor log files could be useful when debugging, but I think better clean them up in the workflow for all kinds of DA runs.

liujake: Maybe commented out the line, not delete it. This way, people can easily add it back if needing debug.

ibanos90: Sounds good. @mos3r3n I think you may want to remove them from the variational and hofx app as well.

mos3r3n: > Sounds good. @mos3r3n I think you may want to remove them from the variational and hofx app as well.

If you agree, I can do that in this PR. @liujake @ibanos90

liujake: fine with me.

ibanos90: Yes, it should be fine

bin/PrepJEDI.csh
```diff
@@ -216,6 +215,11 @@ mkdir -p ${InDBDir}
 rm -r ${OutDBDir}
 mkdir -p ${OutDBDir}
 
+if ( "$ArgAppType" == enkf ) then
+  rm -r ${AnaDBDir}
+  mkdir -p ${AnaDBDir}
```
ibanos90: It is still not clear to me why do we need the `AnaDBDir` folder. I just ran the `getkf_OIE120km_WarmStart` scenario and the `dbOut` folder (where we usually keep all obsout diagnostics) is empty while `dbAna` contains the obsout_da files. Wouldn't this be an issue for the graphics package later? Should we move those files to `dbOut`?

mos3r3n: I store the HofXs of ensemble analyses in dbAna because it is calculated separately from the HofXs of the ensemble background. In that case, if I still save obsout files in dbOut, it will overwrite those HofXs of background.

ibanos90: Would there be a way to merge the files, if that makes sense?

mos3r3n: > Would there be a way to merge the files, if that makes sense?

We can do it with a python script. However, one important thing is that we need to make sure the final output observation number of bak and ana is exactly the same. Also, both HofXs of background and analysis should use the same number of cores.

mos3r3n: The dbOut directory was empty is because 'retainObsFeedback' was set to False. There is inconsistency for HofXs of bak and ana. I will double-check it.

mos3r3n: Now if we set retainObsFeedback to False, even if diagOMA is set to True, it will not execute.

bin/EnKF.csh
mos3r3n: > > Sounds good. @mos3r3n I think you may want to remove them from the variational and hofx app as well.
> 
> If you agree, I can do that in this PR. @liujake @ibanos90

Already made corresponding modifications to HofXs and Variational DA.

ibanos90:
Hi @mos3r3n, I thought a bit more about your changes for the static BC for EnKF and I was wondering if in order to make it a somewhat more generic we could add a variable that indicates if bias correction is `variational` or `static` instead of making it application dependent. What do you think?

Also, now that we have a folder named `biasStatic`, would it make sense renaming the current `bias` => `biasVariational`?

bin/PrepJEDI.csh
ibanos90: After our offline talk, I think we can leave these changes as they are for now and work on a following PR for corresponding changes for merging these files for verification purposes. The goal is that we can be able to produce OmA/OmB stats and plot them together similar to what we do for variational.

scenarios/getkf_OIE60km_WarmStart_BC.yaml
```diff
@@ -0,0 +1,50 @@
...
+        hofx:
+          abi_g16: /glade/campaign/mmm/parc/liuz/pandac_common/obs/ioda_v3/2018/ABIASR/IODANC_THIN15KM_SUPEROB15X15_no-bias-correct
+          abi-clr_g16: /glade/campaign/mmm/parc/liuz/pandac_common/obs/ioda_v3/2018/ABIASR/IODANC_THIN15KM_SUPEROB15X15_no-bias-correct
+          ahi_himawari8: /glade/campaign/mmm/parc/liuz/pandac_common/obs/ioda_v3/2018/AHIASR/IODANC_SUPEROB15X15_no-bias-correct
+          ahi-clr_himawari8: /glade/campaign/mmm/parc/liuz/pandac_common/obs/ioda_v3/2018/AHIASR/IODANC_SUPEROB15X15_no-bias-correct
+      IODASuperObGrid:
+        abi_g16: 15X15
+        ahi_himawari8: 15X15
+rtpp:
+  relaxationFactor: 0.0
```
ibanos90: As we just talked, this section for `rtpp` can be removed from the scenario given that 0. is the default value.

scenarios/letkf_OIE60km_WarmStart.yaml
```diff
@@ -39,7 +39,7 @@ observations:
         abi_g16: 15X15
         ahi_himawari8: 15X15
 rtpp:
-  relaxationFactor: 0.8
+  relaxationFactor: 0.0
```
ibanos90: Similar here

test/testinput/getkf_OIE120km_WarmStart.yaml
```diff
@@ -28,7 +30,7 @@ model:
 observations:
   resource: PANDACArchive
 rtpp:
-  relaxationFactor: 0.8
+  relaxationFactor: 0.0
```
ibanos90: Similar here

test/testinput/letkf_OIE120km_WarmStart.yaml
```diff
@@ -28,7 +28,7 @@ model:
 observations:
   resource: PANDACArchive
 rtpp:
-  relaxationFactor: 0.8
+  relaxationFactor: 0.0
```
ibanos90: Similar here

mos3r3n:
> Hi @mos3r3n, I thought a bit more about your changes for the static BC for EnKF and I was wondering if in order to make it a somewhat more generic we could add a variable that indicates if bias correction is `variational` or `static` instead of making it application dependent. What do you think?
> 
> Also, now that we have a folder named `biasStatic`, would it make sense renaming the current `bias` => `biasVariational`?

I did the modification to directly set the bias coefficient files in the yaml, not linking them to dbOut or dbIn. That would be much cleaner. For the potential variational DA that doesn't update bias coefficient files, I prefer to leave it to the future. I will not rename 'bias' to 'biasVariational' in this PR unless you all think it is necessary. Thanks!

bin/PrepJEDI.csh
mos3r3n: > Would there be a way to merge the files, if that makes sense?

As we discussed before, I will leave it to a separate PR, which focuses more on the ensemble verification in the observation space regarding the issue #215. @ibanos90

ibanos90: Sounds good. Thanks!

ibanos90:
> > Hi @mos3r3n, I thought a bit more about your changes for the static BC for EnKF and I was wondering if in order to make it a somewhat more generic we could add a variable that indicates if bias correction is `variational` or `static` instead of making it application dependent. What do you think?
> > Also, now that we have a folder named `biasStatic`, would it make sense renaming the current `bias` => `biasVariational`?
> 
> I did the modification to directly set the bias coefficient files in the yaml, not linking them to dbOut or dbIn. That would be much cleaner. For the potential variational DA that doesn't update bias coefficient files, I prefer to leave it to the future. I will not rename 'bias' to 'biasVariational' in this PR unless you all think it is necessary. Thanks!

Sounds good, thanks for considering these changes!

mos3r3n:
> > > Hi @mos3r3n, I thought a bit more about your changes for the static BC for EnKF and I was wondering if in order to make it a somewhat more generic we could add a variable that indicates if bias correction is `variational` or `static` instead of making it application dependent. What do you think?
> > > Also, now that we have a folder named `biasStatic`, would it make sense renaming the current `bias` => `biasVariational`?
> > 
> > 
> > I did the modification to directly set the bias coefficient files in the yaml, not linking them to dbOut or dbIn. That would be much cleaner. For the potential variational DA that doesn't update bias coefficient files, I prefer to leave it to the future. I will not rename 'bias' to 'biasVariational' in this PR unless you all think it is necessary. Thanks!
> 
> Sounds good, thanks for considering these changes!

The calculation of HofXs of ensemble analyses needs to modify dbOut to dbAna. When the bias coefficient files are set in the yaml, their locations will be changed to dbAna, too, and diagOMA will fail. I have modified it to store the HofXs of analyses in dbOut but with different suffixes, so dbAna is no longer needed.

initialize/applications/EnKF.py
```diff
@@ -74,6 +85,9 @@ class EnKF(Component):
     # OPTIONS: False (not enabled yet)
     'biasCorrection': [False, bool],
 
+    # directories that stores varBC coefficients that were generated by VarBC
+    'staticVarBcDir': ['directoryToStoreVarBcCoefficients', str],
```
ibanos90: I wonder if we should specify a default path for this variable.

bin/PrepJEDI.csh
```diff
@@ -276,7 +275,11 @@ cd ${WorkDir}
 if ( ${thisValidDate} == ${nextFirstCycleDate} ) then
   set biasCorrectionDir = $initialVARBCcoeff
 else
-  set biasCorrectionDir = ${DAWorkDir}/$prevValidDate/dbOut
+  set biasCorrectionDir = ${DAWorkDir}/$prevValidDate/$OutDBDir
+endif
+# For EnKF, offline bias correction can be applied for bias correction
+if ( "$ArgAppType" == enkf ) then
+  set biasCorrectionDir = ${staticVarBcDir}/${thisValidDate}/${OutDBDir}
```
ibanos90, line 282: It seems like this will work for the case that you use the satbias files from a pre-updated ones in a cycling experiments, but it may not if we use fix ones such as those in `initialVARBCcoeff` path. Should we consider this case as well?

mos3r3n: > It seems like this will work for the case that you use the satbias files from a pre-updated ones in a cycling experiments, but it may not if we use fix ones such as those in `initialVARBCcoeff` path. Should we consider this case as well?

Could make it direct to the initial one if not avaiable.

mos3r3n: As you can see from the lines 342-343, if the satbias files are not available, it will direct to the initial one. In that case, even the staticVarBcDir was not well set, it will make sure the bias correction could work for EnKF.

ibanos90: You are right, lines 342-343 take care of that part. Thanks for pointing that out!

Would specify a default path for `staticVarBcDir` make sense just to ensure this variable is pre-set as intended (in case we intend to use pre-updated satbias files as default) as I mentioned in my other comment?

mos3r3n: Sounds good.

ibanos90 (approved):
Thanks for addressing my comments @mos3r3n!

mos3r3n:
I have updated this branch to the latest and combined EnKF.csh and EnKFObserver.csh into a single script to do both jobs.
We could merge this PR if it is ready.

scenarios/getkf_OIE60km_WarmStart_BC.yaml
```diff
@@ -0,0 +1,48 @@
...
+model:
+  outerMesh: 60km
+  # TODO: make inner and ensemble meshes unnecessary
+  # related to {{PrepareExternalAnalysisInner}} and {{PrepareExternalAnalysisEnsemble}}
+  innerMesh: 60km
+  ensembleMesh: 60km
+observations:
+  resource: PANDACArchiveForVarBC
+  resources:
+    PANDACArchive:
```
ibanos90: Hi @mos3r3n, just noted that here this has to be `PANDACArchiveForVarBC` if you want the observations resources to use your specified paths.

PR #339, opened 2025-01-10
Adapting jcsda-internal/mpas-jedi#1011

byoung-joo:
### Description

This PR adapts the changes required for jcsda-internal/mpas-jedi#1011.

* This PR changes the **wind analysis variable names** (which were MPAS Model's name) to JEDI's generic names (which are ccpp names). 
  * `uReconstructZonal` --> `eastward_wind`
  * `uReconstructMeridional` --> `northward_wind`.
* Note that we keep using the MPAS wind variable names in **state variables**.
* Testing is under way (please see https://github.com/JCSDA-internal/mpas-jedi/pull/1037 ).

#### Tier 1:
 - [x] eda_OIE120km_WarmStart

#### Tier 2 (optional):
 - [ ] GenerateGFSAnalyses
 - [ ] GenerateObs

### DIFF
--- a/config/mpas/variables.csh
+++ b/config/mpas/variables.csh
@@ -10,11 +10,15 @@ set StandardAnalysisVariables = ( \
   spechum \
   surface_pressure \
   temperature \
-  uReconstructMeridional \
-  uReconstructZonal \
+  eastward_wind \
+  northward_wind \
 )
 set StandardStateVariables = ( \
-  $StandardAnalysisVariables \
+  spechum \
+  surface_pressure \
+  temperature \
+  uReconstructMeridional \
+  uReconstructZonal \
   theta \
   rho \
   u \
--- a/config/mpas/variables.csh
+++ b/config/mpas/variables.csh
@@ -10,8 +10,8 @@ set StandardAnalysisVariables = ( \
   spechum \
   surface_pressure \
   temperature \
-  eastward_wind \
   northward_wind \
+  eastward_wind \
 )
 set StandardStateVariables = ( \
   spechum \
--- a/initialize/framework/Build.py
+++ b/initialize/framework/Build.py
@@ -44,7 +44,7 @@ class Build(Component):
         self.variablesWithDefaults['mpas bundle'] = [config._bundle_dir, str]
       else:
         self.variablesWithDefaults['mpas bundle'] = \
-          ['/glade/campaign/mmm/parc/ivette/pandac/codeBuild/mpasBundle_16Dec2024/build_SP', str] ## develop
+          ['/glade/derecho/scratch/bjung/panda-c/code_spectralWindNaming/mpas-bundle/build_single', str] ## develop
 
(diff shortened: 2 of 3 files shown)


config/mpas/variables.csh
```diff
@@ -10,11 +10,15 @@ set StandardAnalysisVariables = ( \
   spechum \
   surface_pressure \
   temperature \
-  uReconstructMeridional \
-  uReconstructZonal \
+  eastward_wind \
```
byoung-joo: This change of order for [u, v] wind variable introduces a small difference in the result. See the comparison of one-week cycling exps. in https://github.com/JCSDA-internal/mpas-jedi/pull/1037#issuecomment-2591067972 .

Should we keep the original order for [u, v] wind ? or alphabetize (as shown now) ?

ibanos90: Thanks for looking into this in depth, @byoung-joo. I think we may want to use the order that gives us no change in the results. But, still no clear to me why the order in which specify the variables would introduce a change.

initialize/framework/Build.py
```diff
@@ -44,7 +44,7 @@ def __init__(self, config:Config, model:Model=None):
         self.variablesWithDefaults['mpas bundle'] = [config._bundle_dir, str]
       else:
         self.variablesWithDefaults['mpas bundle'] = \
-          ['/glade/campaign/mmm/parc/ivette/pandac/codeBuild/mpasBundle_16Dec2024/build_SP', str] ## develop
+          ['/glade/derecho/scratch/bjung/panda-c/code_spectralWindNaming/mpas-bundle/build_single', str] ## develop
```
ibanos90, line 47: Should we move the built code to a more permanent location?

config/mpas/variables.csh
byoung-joo: > I think we may want to use the order that gives us no change in the results. But, still no clear to me why the order in which specify the variables would introduce a change.

I agree, it is still under discussion in mpas-jedi#1037. I am waiting for JCSDA team's response.

initialize/framework/Build.py
byoung-joo: Sure, I will do that at the time of merging the mpas-jedi PR!

ibanos90: Thanks!

config/mpas/variables.csh
byoung-joo: @ibanos90, I'd like to confirm your opinion that `northward_wind` comes first in this line ... to avoid the _unwanted_ change in the result (due to a numerical rounding off error).

ibanos90: Yes, I think we should go ahead with that choice. Thanks for confirming, @byoung-joo!

byoung-joo: Sounds good!

byoung-joo:
This PR will be a part of https://github.com/NCAR/MPAS-Workflow/pull/346.

PR #340, opened 2025-01-21
Modify for suite GenerateExternalAnalyses

zhumingying:
This PR addresses the GenerateGFSAnalyses scenario for MPAS-Workflow version 3.0 and later, which uses MPAS model version 8.2 and replaces static files with the Invariant file. Therefore, InvariantStream information needs to be added to the GenerateExternalAnalyses suite file.

Files modified: initialize/suites/GenerateExternalAnalyses.py
                          scenarios/GenerateGFSAnalyses.yaml
                         

#### Tier 2 (optional):
 - [ ] GenerateGFSAnalyses

### DIFF
--- a/scenarios/GenerateGFSAnalyses.yaml
+++ b/scenarios/GenerateGFSAnalyses.yaml
@@ -2,12 +2,15 @@ suite: GenerateExternalAnalyses
 
 externalanalyses:
   resource: "GFS.RDA"
+  #resource: "GFS.NCEPFTP"
 experiment:
   name: 'GenerateGFSAnalyses'
   prefix: ''
+model:
+  outerMesh: 120km
 hpc:
-  CriticalQueue: economy
-  NonCriticalQueue: economy
+  CriticalQueue: main
+  NonCriticalQueue: main
 workflow:
   first cycle point: 20220606T00
   final cycle point: 20220608T00
--- a/scenarios/GenerateGFSAnalyses.yaml
+++ b/scenarios/GenerateGFSAnalyses.yaml
@@ -9,8 +9,10 @@ experiment:
 model:
   outerMesh: 120km
 hpc:
-  CriticalQueue: main
-  NonCriticalQueue: main
+  CriticalQueue: economy
+  NonCriticalQueue: economy
+  #CriticalQueue: main
+  #NonCriticalQueue: main
 workflow:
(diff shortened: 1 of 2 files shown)


mrixlam:
For the HPC queue section of the code, I think it we would go with "main" queue but specify "job priority" to economy. This would reduce 30% core hour consumption. 
This page could be relevant here: https://ncar-hpc-docs.readthedocs.io/en/latest/pbs/charging/

ibanos90:
Hi @zhumingying, can you add reviewers for this PR and also update the PR's description and tests conducted? Thanks!

scenarios/GenerateGFSAnalyses.yaml
```diff
@@ -2,12 +2,15 @@ suite: GenerateExternalAnalyses
...
 experiment:
   name: 'GenerateGFSAnalyses'
   prefix: ''
+model:
+  outerMesh: 120km
 hpc:
-  CriticalQueue: economy
-  NonCriticalQueue: economy
+  CriticalQueue: main
+  NonCriticalQueue: main
```
ibanos90: Here we can leave it as `economy` as in Derecho `main` will be the default queue if not specified. The way that it works for Derecho is that the variables `CriticalQueue` and `NonCriticalQueue` are used towards the job priority. If you specify `main` as the queue then it uses `regular` as the job priority which costs more as Rubaiat pointed out. To better understand this you can take a look at the function `__convertQueue` in [config.py](https://github.com/NCAR/MPAS-Workflow/blob/develop/initialize/config/Config.py) and also [HPC.py](https://github.com/NCAR/MPAS-Workflow/blob/develop/initialize/framework/HPC.py).

zhumingying: I changed back to economy

ibanos90 (approved):
Thanks for adding this change @zhumingying! I tested and it worked correctly. You can check the box for the test in the PR's description.

byoung-joo:
I was wondering if we already includes `Invariant` stream in the workflow, but realized that it was missing in this `GenerateExternalAnalyses` suite. Nice addition, @zhumingying !

junmeiban (approved):
Many thanks @zhumingying !

zhumingying:
Thanks all!

PR #344, opened 2025-01-29
Change tools/create_amb_in_nc.py to create cdf5 data files.

jim-p-w:
It currently creates netcdf4 files, which aren't readable by the SMIOL I/O library.

Change the io_type of the iau stream to be pnetcdf,cdf5 in config/mpas/forecast/streams.atmosphere.

#### Tier 1:
 - [x] eda_OIE120km_WarmStart

#### Tier 2 (optional):
 - [ ] GenerateGFSAnalyses
 - [ ] GenerateObs

### DIFF
--- a/tools/create_amb_in_nc.py
+++ b/tools/create_amb_in_nc.py
@@ -43,7 +43,9 @@ def create_anal_incr_to_nc():
     file_a = 'an.%s.nc'%(tanl.strftime('%Y-%m-%d_%H.%M.%S'))	# Input analysis
     file_b = 'bg.%s.nc'%(tanl.strftime('%Y-%m-%d_%H.%M.%S'))	# Input background
     famb =  'AmB.%s.nc'%(tiau.strftime('%Y-%m-%d_%H.%M.%S')) 	# Output analysis increments
-    fout  = Dataset(famb, 'w')
+    fout  = Dataset(famb, 'w', format="NETCDF3_64BIT_DATA")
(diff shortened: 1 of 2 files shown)


tools/create_amb_in_nc.py
```diff
@@ -58,7 +60,7 @@ def create_anal_incr_to_nc():
     fout.input_xtime = tanl.strftime('%Y-%m-%d_%H:%M:%S') + ' UTC'
     fout.IAU_window_length_s = args.IAU_window_length_s
     fout.input_interval = "initial_only"
-    fout.io_type = "netcdf4"
+    fout.io_type = "NETCDF3_64BIT_DATA" #cdf5
```
liujake, line 63: I am not sure this is the right io_type name to use. 'NETCDF3' is susperious to me (older version 3?) and 64BIT seems to be for double precision, while we are using single precision model build. @amstokely or michael may know if this is the one to use.

jim-p-w: @liujake 
[Here are the netcdf docs](https://unidata.github.io/netcdf4-python/#creatingopeningclosing-a-netcdf-file)
ncdump on the file shows
```
ncdump -k AmB.2018-04-14_21.00.00.nc
cdf5
```

liujake: Ok. Looks like 64BIT is for the larger file size, not precision.

liujake (approved):

PR #356, opened 2025-02-21
Enable running the Forecast task on GPU nodes on derecho.

jim-p-w:
### Description
This PR enables running the forecast step on derecho GPU nodes.

To run the forecast on gpu nodes
1. provide the build directory for the GPU build of mpas_atmosphere in the top level scenario yaml:
```
    build:
      forecast directory: path-for-mpas-model-gpu-build
```
2. add the GPUPerNode attribute of the forecast job:
```
    forecast:
      job:
        GPUPerNode: 4
```

Task.py will change the select PBS directive to send a job to the gpu queue if the task has a `GPUPerNode` attribute.

#### Tier 1:

I ran `test/testinput/3dvar_OIE120km_WarmStart.yaml` both with specifying to use GPU nodes and without specifying to use GPU nodes.

I verified that the Forecast1 step was submitted to the gpu queue when
  1.  `GPUPerNode` was specified with a value of `4` in `test/testinput/3dvar_OIE120km_WarmStart.yaml` and 
  2.  `forecast directory` was specified in `test/testinput/3dvar_OIE120km_WarmStart.yaml`

I verified the Forecast1 step was submitted to the cpu queue when
  1.  `GPUPerNode` was specified with a value of `0` in `test/testinput/3dvar_OIE120km_WarmStart.yaml` and 
  2.  `forecast directory` was *not* specified in `test/testinput/3dvar_OIE120km_WarmStart.yaml`

I verified that the test scenario ran to successful completion regardless of the type of nodes the Forecast1 step ran on.

### Notes
1. The GPU port of the MPAS-Model dynamical core is in process. At this time it runs correctly on the GPU nodes, but performance is an issue and will remain so until the GPU port is completed
2. An updated README with instructions to get your scenario to run the forecast steps on gpu's will be in a separate PR.

### DIFF
--- a/test/testinput/3dvar_OIE120km_WarmStart.yaml
+++ b/test/testinput/3dvar_OIE120km_WarmStart.yaml
@@ -7,6 +7,10 @@ firstbackground:
 forecast:
   # turn off post to reduce overhead
   post: []
+  job:
+    # to run forecast on gpu's, set GPUPerNode to 4 and
+    # uncomment forecast directory below
+    GPUPerNode: 0
 hpc:
   CriticalQueue: economy
   NonCriticalQueue: economy
@@ -26,3 +30,5 @@ variational:
 workflow:
   first cycle point: 20180414T18
   final cycle point: 20180415T06
+build:
+  #forecast directory: /glade/derecho/scratch/jwittig/repos-s/mpas-bundle-io/build-nvhpc
--- a/bin/Forecast.csh
+++ b/bin/Forecast.csh
@@ -363,6 +363,7 @@ else
     if ( -e $f ) rm -v $f
   end
   ln -sfv ${ForecastBuildDir}/${ForecastEXE} ./
+  echo "{mpiCommand} ${mpiCommand} ./${ForecastEXE}"
   ${mpiCommand} ./${ForecastEXE}
 
 
--- a/config/environmentForecast.csh
+++ b/config/environmentForecast.csh
@@ -31,20 +31,40 @@ else
     else if ("$NCAR_HOST" == "derecho") then
       source /etc/profile.d/z00_modules.csh
       module purge
+      module load ncarenv/23.09
+      module reset
 
-      module load intel-oneapi/2023.0.0
-      module load cray-mpich/8.1.25
-      module load parallel-netcdf/1.12.3
-      module load parallelio/2.5.10
-      module load ncl/6.6.2
-      module load hdf5/1.12.2
-      module load netcdf/4.9.2
-      module load ncarcompilers/1.0.0
-      module load nco/5.1.4
-      setenv mpiCommand mpiexec
+      if ("$NGPUS" == "0") then
+        module load intel-oneapi/2023.0.0
+        module load cray-mpich/8.1.25
+        module load parallel-netcdf/1.12.3
+        module load parallelio/2.5.10
+        module load ncl/6.6.2
+        module load hdf5/1.12.2
+        module load netcdf/4.9.2
+        module load ncarcompilers/1.0.0
+        module load nco/5.1.4
+        setenv mpiCommand mpiexec
+      else if("$NGPUS" == "4") then
+        module load cuda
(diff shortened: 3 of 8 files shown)


jim-p-w:
The changes to Build.py are to support building `atmosphere_model` directly via make in an MPAS-Model repo clone, or to build `mpas_atmosphere` using cmake in an mpas-bundle environment.

test/testinput/3dvar_OIE120km_WarmStart.yaml
```diff
@@ -26,3 +26,5 @@ variational:
 workflow:
   first cycle point: 20180414T18
   final cycle point: 20180415T06
+build:
+    #forecast directory: /glade/derecho/scratch/jwittig/repos-s/mpas-bundle-io/build-nvhpc
```
ibanos90: I wonder if these should be added to the test `ForecastFromGFSAnalysesMPT` that is specifically to test the forecast suite.

scenarios/defaults/forecast.yaml
```diff
@@ -6,6 +6,13 @@ forecast:
       PEPerNode: 128
       baseSeconds: 60
       secondsPerForecastHR: 500
+      # uncomment the following to run the forecast on a gpu node on derecho
+      #GPUPerNode: 4
```
ibanos90: I wonder if there may be another way to add this but not requiring to have to comment or uncomment this option

test/testinput/3dvar_OIE120km_WarmStart.yaml
jim-p-w: Are you suggesting that I should run this scenario w/ a gpu build? I can do that.

scenarios/defaults/forecast.yaml
jim-p-w: That was my intent when I initially required passing `-g` to Run.py.
I could change things to ignore `GPUPerNode` if the value is `0`.

ibanos90: That sounds good to me. Then we just need to specify `GPUPerNode: 4` under `job` in the scenario yaml file, correct?

test/testinput/3dvar_OIE120km_WarmStart.yaml
ibanos90: Well, my thoughts were that if we add the gpu built code to the `ForecastFromGFSAnalysesMPT` test then it will be tested routinely. Also, it will need the configuration for the gpu per node then we have a reference of how to use it.

scenarios/defaults/forecast.yaml
jim-p-w: We'll also need to specify the `forecast directory`

test/testinput/3dvar_OIE120km_WarmStart.yaml
jim-p-w: Sounds good. I will do that.

scenarios/defaults/forecast.yaml
jim-p-w: I'll make the change.

test/testinput/3dvar_OIE120km_WarmStart.yaml
jim-p-w: Done.

scenarios/defaults/forecast.yaml
jim-p-w: Done.

ibanos90 (approved):
Thanks a lot for adding this capability, @jim-p-w! I ran the tier 1 and all test succeeded. I also ran the `3dvar_OIE120km_WarmStart` test with the gpu queue and gpu build activated and worked as expected!

jim-p-w:
> Thanks a lot for adding this capability, @jim-p-w! I ran the tier 1 and all test succeeded. I also ran the `3dvar_OIE120km_WarmStart` test with the gpu queue and gpu build activated and worked as expected!

@ibanos90 Thanks for taking the time to test this!

PR #362, opened 2025-04-02
Add a scenario for running the weekly cylc job via cron.

jim-p-w:
This is the same scenario as scenarios/3denvar_OIE120km_WarmStart_VarBC.yaml, but the end point is 20180421T18 and it uses the economy queue.

### Tests completed
I verified the weekly cron job is running this scenario, and the scenario is completing successfully.


ibanos90 (approved):

ibanos90:
I think you can remove the Tier 1 and 2 from the PR's description

PR #365, opened 2025-04-20
Using MPAS-JEDI's own executable to calculate ensemble mean and variance

mos3r3n:
### Description
Previously, the ensemble mean and variance were calculated using an external executable. This PR uses the mpasjedi_ens_mean_variance.x within MPAS_JEDI to do it.

### Issue closed

Closes #(if applicable)

#### Tier 1:
 - [x] eda_OIE120km_WarmStart

#### Tier 2 (optional):
 - [ ] GenerateGFSAnalyses
 - [ ] GenerateObs


config/mpas/variables.csh
```diff
@@ -10,7 +10,14 @@ set MPASHydroIncrementVariables = ( \
   rain_water \
   snow_water \
 )
-set MPASHydroStateVariables = (${MPASHydroIncrementVariables} cldfrac)
```
jim-p-w, line 13: When I ran `test/testinput/eda_OIE120km_WarmStart.yaml` the HofX steps all failed with
`mpasjedi_vc_model2geovars::changevar: cldfrac must be added to the state  variables in order to populate the var_cldfrac geovar with the MPAS diagnostic cloud  fraction`

I reverted the change to `config/mpas/variables.csh` and the test case passed.
I also reran `test/testinput/letkf_OIE120km_WarmStart.yaml` with the reverted variables.csh and it passed.

ibanos90: I get the same for several tests on the `HofX` step.

mos3r3n: I used it to also calculate the ensemble mean of cldfrac. Without  specified in state variables, the mean of cloud_area_fraction_in_atmosphere_layer could not be calculated. Maybe we need to add both? I am not sure but will try it later.

byoung-joo: I am a bit confused. The model variable name `cldfrac` has not changed as a generic name yet (i.e., `cloud_area_fraction_in_atmosphere_layer`).  It is also the same for `effective_radius_of_xxxx` variables. All these variable should go through the "variable change". I think this PR would not work with `develop` branch of `mpas-jedi`.   @mos3r3n, I wonder if this PR is compatible with some of your feature branch, not with `develop`.

mos3r3n: It is from the develop branch. I thought those variables were general variables already. I will make a commit to correct them back.

byoung-joo: You can check the list of variables here: https://docs.google.com/spreadsheets/d/1Kh_Ypp8Uwm69yfIhGcAlijgmzU7iY_ZpnBS9SQZoQTg/edit?gid=0#gid=0

mos3r3n: Thanks, BJ!
I modified them to cldfrac, re_cloud, re_ice, and re_snow. Now it works.

ibanos90: Great, I tested and it HoxF works now.

ibanos90 (approved):

config/mpas/variables.csh
```diff
@@ -10,7 +10,14 @@ set MPASHydroIncrementVariables = ( \
...
+
+set MPASHydroStateVariables = ( \
+  $MPASHydroIncrementVariables \
+  cldfrac \
+  re_cloud \
+  re_snow \
+  re_ice \
```
byoung-joo, line 19: @mos3r3n , I'd like to confirm that the results are the same (for example, ensMeanVariance) even if we remove these three variables, `re_cloud`, `re_snow`, and `re_ice`.

To clarify the procedure, the list of state variables should come from yaml config. This is not the case in current mpas-jedi's develop branch. See for example https://github.com/JCSDA-internal/mpas-jedi/blob/develop/src/mpasjedi/State/mpas_state_interface_mod.F90#L70-L94 .
This was discussed with @mos3r3n and @jim-p-w, separately before.

The usecase of mpas-jedi is getting more complex, so it would be nice to make this part clear in near future.

config/mpas/variables.csh
```diff
@@ -24,6 +31,7 @@ set StandardStateVariables = ( \
   air_potential_temperature \
   dry_air_density \
   u \
+  w \
```
byoung-joo, line 34: Is this a necessary change ?

config/mpas/variables.csh
```diff
@@ -36,6 +44,8 @@ set StandardStateVariables = ( \
   vegetation_area_fraction \
   eastward_wind_at_10m \
   northward_wind_at_10m \
+  air_temperature_at_2m \
+  water_vapor_mixing_ratio_wrt_moist_air_at_2m \
```
byoung-joo: Is this a necessary change ?

mos3r3n: I think for future surface assimilation purpose, these two variables could be added.

config/mpas/variables.csh
mos3r3n: If these variables are not included, their mean will not be calculated. For all-sky HofX of the ensemble mean, these variables should be available.
I haven't tested the DA applications. Maybe we need to tested via 1-week cycling test @jim-p-w .

config/mpas/variables.csh
mos3r3n: I'm not quite sure whether the w variable is a necessity to initialize the model. 
If in the future we assimilate radial velocity data, this variable should be added to the state variables.

byoung-joo:
In my comments above, I am trying to keep the necessary changes that are relevant to this PR in general.

Additional features can be done in separate PRs. This will also help us keep track of changes in this repository.

mos3r3n:
> In my comments above, I am trying to keep the necessary changes that are relevant to this PR in general.
> 
> Additional features can be done in separate PRs. This will also help us keep track of changes in this repository.

Thanks @byoung-joo! I agree that T2 and Q2 are not used in DA or prognostic variables of MPAS, so I will not add them in this PR. The w wind is a prognostic variable of MPAS, so I will keep it in this PR.

byoung-joo:
> The w wind is a prognostic variable of MPAS, so I will keep it in this PR.

Sound good to me. This application seems to be unique. It reads the ensemble using "background" stream and writes out using "background" stream.  `w` is already included in our `stream_list.atmosphere.background`.

I think that adding `w` to state variabe list would not affect the existing variational application.

config/mpas/variables.csh
byoung-joo: I realized that the arithmetic operations only work with [`vars_ci`](https://github.com/JCSDA-internal/mpas-jedi/blob/develop/src/mpasjedi/Fields/mpas_fields_mod.F90#L237), which comes from the yaml directly. So this is a necessary change for `mpasjedi_ens_mean_variance.x`.

byoung-joo:
I confirmed that adding `w` to `StandardAnalysisVariables` does not affect the existing Variational application.
I have not tested the effect of adding `re_cloud`, `re_ice`, and `re_snow` to `MPASHydroStateVariables`, but in my understanding this will not affect the existing Variational application.

byoung-joo (approved):

PR #404, opened 2025-12-05
Use the previous cycle's sea surface analysis file when running IAU.

### LINKED ISSUE #353: The 3denvar_OIE120km_IAU_WarmStart test fails when using SMIOL
test/testinput/3denvar_OIE120km_IAU_WarmStart.yaml fails in the Forecast1 step when using mpas_atmosphere built with the SMIOL library.
The test passes when MPAS-Model is built with the PIO I/O library.

**To Reproduce**
1. Run the test in an environment which uses an mpas-bundle build made with the MPAS-Model SMIOL library
(the develop branch of mpas bundle post Jan 30 2024 will default to MPAS-Model using the SMIOL I/O library.)

The test will fail in the Forecast1 step.
mpas_atmosphere fails to read elements of the `x1.40962.sfc_update.nc` data file. The file only has entries dated 2018-04-15_00.00.00.
The test's namelist.atmosphere is configured with `config_start_time = '2018-04-14_21:00:00'`.

Adding the attribute `reference_time="2018-04-14_21:00:00"` to the definition of the "surface" stream in the streams.atmosphere file had no effect.
Changing the attribute `input_interval` from `"initial_only"` to `"24:00:00"` in the definition of the "surface" stream in the streams.atmosphere file had no effect.

Changing a parameter in the call to `MPAS_stream_mgr_read` from `atm_core_run` ( mpas_atm_core.F ) fixed this problem (but is not a suitable fix). The fix was to change the parameter defining the time to read from `MPAS_STREAM_LATEST_BEFORE` to `MPAS_STREAM_NEAREST`
With MPAS_STREAM_LATEST_BEFORE only enties up to the curent valid time are read. Using MPAS_STREAM_NEAREST enables reading entries after the current valid time (which is usually not desirable).

jim-p-w:
Re-enable the 3denvar_OIE120km_IAU_WarmStart test in test1.yaml

### Description
This PR fixes an mpas-atmosphere crash when using IAU while generating a forecast.
The initial time for running the model is set to t0 - 3hr when running IAU (instead of t0).
mpas-atmosphere will not read data from a file with a time stamp which is later than the current time, so
the sea surface analysis file from the previous cycle's run must be used (instead of the current cycle's).

### Issue closed
Closes #353

#### Tier 1:
 - [x] eda_OIE120km_WarmStart

#### Tier 2 (optional):
 - [ ] GenerateGFSAnalyses
 - [ ] GenerateObs

### DIFF
--- a/bin/Forecast.csh
+++ b/bin/Forecast.csh
@@ -241,14 +241,31 @@ if ("${ArgUpdateSea}" == True) then
   else
     # otherwise use deterministic analysis for all members
     # 60km and 120km
-    setenv SeaAnaDir ${deterministicSeaAnaDir}
+
+    # IAU starts the model before the 'current' analysis time, so we need to get
+    # the previous analysis for sea surface fields
+    if ( ${self_IAU} == True ) then
+      set lastEADir = ${ExperimentDirectory}/`echo "${ExternalAnalysesDirOuter}" \
+        | sed 's@{{thisValidDate}}@'${prevCycleDate}'@' \
+        `
+      setenv SeaAnaDir ${lastEADir}
+      setenv SeaFilePrefix ${deterministicSeaFilePrefix}.${prevMPASFileDate}
+    else
+      setenv SeaAnaDir ${deterministicSeaAnaDir}
+      setenv SeaFilePrefix ${deterministicSeaFilePrefix}
+    endif
     setenv seaMemFmt "${deterministicSeaMemFmt}"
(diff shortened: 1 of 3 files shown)


ibanos90 (approved):
Thanks for fixing this issue @jim-p-w! Your changes look and I ran the test with IAU and it worked correctly.

byoung-joo (approved):
Thank you for your effort, @jim-p-w !

PR #414, opened 2026-01-14
Use a mutable stream, instead of immutable stream `da_state`

### LINKED ISSUE #412: Make the list of variables in the background file flexible
In the 2-stream I/O for cycling DA, the list of variables for background file is defined as immutable stream `da_state`, which can be found under https://github.com/MPAS-Dev/MPAS-Model/blob/master/src/core_atmosphere/Registry.xml#L1197 . In current procedure, the immutable stream `da_state` is only being used for "output", NOT for "input".

If we want to add an additional variable to the background file (i.e., `da_state` stream), this requires the code change in the MPAS-Model repository. Changing the list of `da_stream` stream does not affect the typical MPAS cold-start forecast capability, but still we need to go though the PR procedure.

Instead, we can use the mutable stream for defining the list of variables for background file, similar to `stream_list.atmosphere.background`, `stream_list.atmosphere.analysis`, `stream_list.atmosphere.ensemble`, and `stream_list.atmosphere.control`.

This requires a minimal change to the existing procedure. Please see the feature branch [feature/configurable_dastate](https://github.com/NCAR/MPAS-Workflow/tree/feature/configurable_dastate) . 

Here, a new text file `stream_list.atmosphere.dastate` is added, which defines the list of variables and the `streams.atmosphere` file is modified accordingly. This way, users can define their own variable list that is needed for their experiment (if the variables are readily available from `Registry.xml` of MPAS-Model). This list should include the minimum required variables ( needed for typical DA and cycling, which means... initializing the next forecast ), in addition to the additional user-desirable variables.

With this, for example, users may add `o3vmr` or aerosol variable to the background file.

If this happens, we should document the role of `dastate` stream and share it with users...

byoung-joo:
### Description
This PR uses the mutable stream `dastate`, instead of immutable stream `da_state`. 
Users can configure the list of variables for the background file with `stream_list.atmosphere.dastate`, without modifying the `Registry.xml` of MPAS-Model. 

FYI: We can not define another mutable stream `da_state` because the name is already defined as immutable stream in the Registry. So `dastate` is being used here.
 
### Issue closed
Closes #412 

#### Tier 1:
 - [x] eda_OIE120km_WarmStart

#### Tier 2 (optional):
 - [ ] GenerateGFSAnalyses
 - [ ] GenerateObs


byoung-joo:
This is ready for review. I confirmed that the results are identical for Tier 1 tests.

config/mpas/forecast/streams.atmosphere
```diff
@@ -14,13 +14,13 @@
                   input_interval="initial_only" />
 
 <immutable_stream name="da_state"
-                  type="output"
+                  type="none"
```
jim-p-w: In section 5.1 of the mpas users guide it states `Note that, for immutable streams, the type attribute may not be changed at run-time.`

Also, given that the `output_interval` is changing to `none`  I assume the value of `filename_template` is immaterial?

byoung-joo: Hi @jim-p-w,

Thanks for your comment, but I forgot to push a commit, which deletes this immutable stream from all streams file.
In future, MPAS Model can delete this immutable stream from `Registry.xml` if we all adapt this approach.

I will push it now.

byoung-joo: I just pushed the missing commit and FYI, my test has included this last commit.

jim-p-w:
Should the variables defined in stream_list.atmosphere.dastate be the same as the variables which used to go into the mpasout file before this change?

Differences between  ./CyclingFC/2018041418/mpasout.2018-04-15_02.00.00.nc from running 3dvar_OIE120km_WarmStart  on Jan 13 with variables defined in config/mpas/forecast/stream_list.atmosphere.dastate shows

variables in mpasout not in config/mpas/forecast/stream_list.atmosphere.dastate:
qc
qg
qi
qr
qs
qv

variables in stream_list.atmosphere.dastate not in mpasout:
isice_lu
iswater_lu
refl10cm

byoung-joo:
> Should the variables defined in stream_list.atmosphere.dastate be the same as the variables which used to go into the mpasout file before this change?

Yes, that is correct.

I have compared, for example,
Test from PR408, using immutable `da_state`
/glade/derecho/scratch/bjung/pandac/bjung_3dvar_OIE120km_ColdStart_TEST_PR408/CyclingFC/2022020100/mpasout.2022-02-01_06.00.00.nc
Test with this PR, using mutable `dastate`
/glade/derecho/scratch/bjung/pandac/bjung_3dvar_OIE120km_ColdStart_TEST_PR414/CyclingFC/2022020100/mpasout.2022-02-01_06.00.00.nc


qx variables are part of `scalars`.
I see the other three variables from my PR408 forecast files. Let me check further...

jim-p-w:
OK the variable list in stream_list.atmosphere.dastate looks good to me.

jim-p-w (approved):
Thanks BJ!

junmeiban (approved):
Tested two cycles by using scenarios/3dhybrid-allsky_O30kmIE60km_SpecifiedEnsemble_VarBC_AMSUA.yaml and ran successfully. Many thanks @byoung-joo

PR #418, opened 2026-02-18
Add getting and converting ERA5 ml analyses to MPAS ICs

ibanos90:
### Description
This PR adds the capability to get and convert ERA5 model level analyses to MPAS format. It relies on [era5_to_int](https://github.com/NCAR/era5_to_int.git) code to get the ERA5 ml data on GDEX and the NetCDF files to WPS intermediate. Then, `mpas_init` is run to create the MPAS init files. A new scenario is added to test this functionality.

### Issue closed
None

### Tests completed
 - [x] GenerateERA5Analyses.yaml


scenarios/defaults/externalanalyses.yaml
```diff
@@ -80,3 +80,15 @@ externalanalyses:
...
+      GDEX:
+        common:
+          UngribPrefix: ERA5
+          PrepareExternalAnalysisTasks:
+          - "GetERA5AnalysisFromGDEX"
+          - "ExternalAnalysisToMPAS-{{mesh}}"
+          - "ExternalAnalysisReady__"
+          job:
+            GetAnalysisFrom:
+              retry: '40*PT10M'
```
jim-p-w: Do we really want to retry 40 times before failing the task? With a 10 minute delay between each retry this could take over 6 hours before failing.

scenarios/GenerateERA5Analyses.yaml
```diff
@@ -0,0 +1,18 @@
...
+model:
+  outerMesh: 120km
+hpc:
+  CriticalPriority: economy
+  NonCriticalPriority: economy
+  #CriticalQueue: main
+  #NonCriticalQueue: main
+workflow:
+  first cycle point: 20220606T00
+  final cycle point: 20220606T06
```
jim-p-w, line 17: FMI How did you determine the number of cycles?

tools/era5_to_int.py
```diff
@@ -0,0 +1,725 @@
...
+
+    glade_paths = [
+        # New Glade paths as of March 2025
+        '/glade/campaign/collections/rda/data/d633006/e5.oper.an.ml/',
+        '/glade/campaign/collections/rda/data/d633000/e5.oper.an.pl/',
+        '/glade/campaign/collections/rda/data/d633000/e5.oper.an.sfc/',
+        '/glade/campaign/collections/rda/data/d633000/e5.oper.invariant/197901/',
+        '/glade/campaign/collections/rda/data/d633006/e5.oper.invariant/',
+
+        # Old Glade paths prior to March 2025
```
jim-p-w: Seems like references to the old rda files can be removed (since they no longer exist).

scenarios/defaults/externalanalyses.yaml
ibanos90: @jim-p-w Thanks for pointing this out! I was following the previous example, but the goal here is different.

scenarios/GenerateERA5Analyses.yaml
ibanos90: That's a good point. This is usually very helpful, as it allows us to have more active points than the default four. Again, I just left what was in the original YAML that I copied this from.

tools/era5_to_int.py
ibanos90: I agree with you. As this script is a copy of the original one here: [era5_to_int.py](https://github.com/NCAR/era5_to_int/blob/main/era5_to_int.py), I kept as it is, but will remove those lines.

bin/GetERA5AnalysisFromGDEX.csh
```diff
@@ -0,0 +1,92 @@
...
+endif
+
+# ================================================================================================
+
+echo "Getting ERA5 analysis from GDEX"
+
+# Link ECMWF coefficients
+ln -sf ${ModelConfigDir}/initic/ecmwf_coeffs ./ecmwf_coeffs
+
+# Convert ERA5 NetCDF to WPS intermediate
```
jim-p-w, line 72: Consider adding the following to [Build/py](https://github.com/NCAR/MPAS-Workflow/blob/686901c46f509e753727253f616b0a6cd62db71b/initialize/framework/Build.py#L176) around line 176
```
self._set('era5_to_int', 'era5_to_int.py')
self._set('era5_to_intBuildDir', '/glade/campaign/mmm/parc/ivette/pandac/converters/era5_to_int')
```
then change ```setenv myCommand `$era5_to_int ${datestr}` ```  to
```
ln -sf $era5_to_intBuildDir/ecmwf_coeffs .
setenv myCommand `$era5_to_intBuildDir/$era5_to_int ${datestr}`
```

Then you can remove `era5_to_int` from tools.csh and remove the files from the era5_to_int repository.

And clone https://github.com/NCAR/era5_to_int to `/glade/campaign/mmm/parc/ivette/pandac/converters/era5_to_int`

ibanos90: That's a great idea @jim-p-w! Will do!

junmeiban (approved):
Many thanks @ibanos90 for adding this function. I generated MPAS ICs from ERA5 ml analyses successfully (1day for 2025 case)

jim-p-w (approved):

ibanos90:
Thanks for your reviews and tests, @jim-p-w and @junmeiban!

initialize/data/ExternalAnalyses.py
```diff
@@ -210,13 +232,13 @@ def export(self, dtOffsets:list=[0]):
       # ungrib
       base = 'UngribExternalAnalysis'
       queue = 'UngribExternalAnalyses'
-      if base in self['PrepareExternalAnalysisOuter']:
+      if base in self['PrepareExternalAnalysisOuter'] and self['externalanalyses__UngribPrefixOuter'] != 'ERA5':
```
jim-p-w, line 235: When I try to run `test/testinput/3dvar_OIE120km_ColdStart.yaml` this line is throwing an exception. It looks like the key `externalanalyses__UngribPrefixOuter` doesn't exist for that scenario.

ibanos90: Thanks for pointing this out @jim-p-w! I should have checked this more carefully. I found a solution by doing `if self.__ungribtask is not None:`  instead. I will add this fix into my open PR.

PR #424, opened 2026-04-17
Use default option for config_extrap_airtemp in namelist.init_atmosphere

ibanos90:
### Description
In [PR#418](https://github.com/NCAR/MPAS-Workflow/pull/418), the capability to convert ERA5 model-level analyses to the MPAS initial conditions format was added. However, when using the current `linear` option in `config/mpas/initic/namelist.init_atmosphere`, unrealistic values of potential temperature (θ) are produced (min=17K), leading to model forecast failure. In contrast, using `lapse-rate` (the default in the User’s Guide) or `constant` both yield physically realistic values. While `linear` performs well for GFS analyses and ERA5 pressure-level data,  it does not properly handle ERA5 model-level data. Therefore, in this PR, the default option is changed to `lapse-rate` as it works across all external data sources. Additionally, a new scenario is introduced to generate forecasts using ERA5 model-level data.

### Issue closed
None

#### Tier 1:

#### Tier 2:
 - [x] ForecastFromGFSAnalyses
 - [x] ForecastFromERAmlAnalyses

### DIFF
--- a/config/mpas/initic/namelist.init_atmosphere
+++ b/config/mpas/initic/namelist.init_atmosphere
@@ -29,7 +29,7 @@
     config_tc_vertical_grid = true
 /
 &interpolation_control
-    config_extrap_airtemp = 'linear'
+    config_extrap_airtemp = 'lapse-rate'
 /
 &preproc_stages
     config_static_interp = false
--- a/scenarios/ForecastFromERAmlAnalyses.yaml
+++ b/scenarios/ForecastFromERAmlAnalyses.yaml
@@ -1,0 +1,40 @@
+suite: ForecastFromExternalAnalyses
+
+externalanalyses:
+  resource: "ERA5.GDEX"
+
+experiment:
+  name: 'ForecastFromERA5mlAnalyses'
+  prefix: ''
+
+extendedforecast:
+  meanTimes: T00,T12
+  lengthHR: 24
+  outIntervalHR: 12
+  #execute: False # uncomment if forecasts are already completed
+
+  # optionally turn on extended forecast post
+  post: [verifyobs, verifymodel]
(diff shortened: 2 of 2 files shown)


config/mpas/initic/namelist.init_atmosphere
```diff
@@ -29,7 +29,7 @@
     config_tc_vertical_grid = true
 /
 &interpolation_control
-    config_extrap_airtemp = 'linear'
+    config_extrap_airtemp = 'lapse-rate'
```
jim-p-w, line 32: Is this going to impact all of the scenarios which run a forecast?

ibanos90: That's a good question @jim-p-w! Scenarios that run `mpas_init_atmosphere` will now be using this option. My test `ForecastFromGFSAnalyses` confirms it runs with no issue. I will run other tests to confirm any differences.

ibanos90: Just confirming the scenarios that run `mpas_init_atmosphere` complete successfully. As can be expected, some differences are found in theta field at level 0 in the init file due to the change in the method used for the extrapolation of air temperature above/below first-guess levels.

jim-p-w: > Just confirming the scenarios that run `mpas_init_atmosphere` complete successfully. As can be expected, some differences are found in theta field at level 0 in the init file due to the change in the method used for the extrapolation of air temperature above/below first-guess levels.

@ibanos90 Thanks for checking this.

jim-p-w (approved):

byoung-joo (approved):

